PR #1, opened 2019-05-29
Optimize table customization. Primarily avoid resolving columns that …

labkey-jeckels:
…don't always exist

Part of the work for Issue 37554: Optimize Linked Schema performance

### DIFF
--- a/ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
+++ b/ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
@@ -71,6 +71,7 @@ import java.io.Writer;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Calendar;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
```diff
@@ -356,45 +228,152 @@ private void doSharedCustomization(AbstractTableInfo ti)
                     UserSchema us = getEHRStudyUserSchema(ti);
                     if (us != null)
                     {
-                        col.setFk(new QueryForeignKey(us, null, "Animal", "Id", "Id"));
+                        col.setFk(new QueryForeignKey(ti.getUserSchema(), ti.getContainerFilter(), us, null, "Animal", "Id", "Id"));
```
labkey-ankurj: would it be better to store ti.getUserSchema() and ti.getContainerFilter() in temp variables outside this for loop?

labkey-jeckels: Simple getters like this are super-fast, so it's not a perf concern. I'd consider this equally good code, so inclined to just keep it as-is unless I'm missing something.

labkey-ankurj: got it, thanks.

PR #9, opened 2019-06-14
Issue 37705: Animal history returning whole database when invalid ID …

labkey-martyp:
…entered

Invalid ID entered in single animal filter caused whole database to be returned in most reports.  This change will validate ID's returned from the ID/Alias check and display "Animal ID not found" message in report window if ID is not valid.

### DIFF
--- a/ehr/resources/web/ehr/panel/AnimalHistoryPanel.js
+++ b/ehr/resources/web/ehr/panel/AnimalHistoryPanel.js
@@ -151,7 +151,7 @@ Ext4.define('EHR.panel.AnimalHistoryPanel', {
 
     filterTypes: [{
         xtype: 'ehr-singleanimalfiltertype',
-        inputValue: EHR.panel.SingleAnimalFilterType.filterName,
+        inputValue: LDK.panel.SingleSubjectFilterType.filterName,
         label: 'Single Animal',
         nounSingular: 'Animal',
         aliasTable: {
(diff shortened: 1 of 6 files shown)


ehr/resources/web/ehr/panel/AnimalHistoryPanel.js
```diff
@@ -82,6 +82,58 @@ Ext4.define('EHR.panel.AnimalHistoryPanel', {
...
+                itemId: 'reportStatus',
+                html: '<div class="alert alert-warning" role="alert">Report Status: <strong>'
+                        + Ext4.util.Format.htmlEncode(tab.report.reportStatus)
+                        + '</strong></div>'
+            });
+
+            tab.add(status);
+        }
+
+        if (tab.report.warning) {
```
labkey-jeckels: From what I can tell, this if statement is the only difference from TabbedReportPanel.displayReport(). Can we consolidate code? Could we replace the TabbedReportPanel implementation with this one? All reports without a warning should be unaffected, right?

bbimber: As noted on the email to Marty I also have concerns and would favor simply trying to make validation asynchronous.  I sent a proposal on email, but we can shift the conversation here is preferred.

labkey-jeckels:
Clean up old PR

PR #12, opened 2019-07-02
Merge from wnprc18.3 svn r. 64002 - 64058

labkey-bpatel:
-Item 5826: [WNPRC] Update DeleteBillingPeriodAction to delete from ehr_billing.invoice, invoiceItems and invoiceRuns with an audit trail.
-Update BillingUtils.isBilling() to make it functional instead of it always returning true.


labkey-bpatel:
wrong branch name, needed to suffix with 19.2

PR #13, opened 2019-07-02
Merge from wnprc18.3 svn r. 64002 - 64108

labkey-bpatel:
-Item 5826: [WNPRC] Update DeleteBillingPeriodAction to delete from ehr_billing.invoice, invoiceItems and invoiceRuns with an audit trail.
-Update BillingUtils.isBilling() to make it functional instead of it always returning true.

### DIFF
--- a/ehr_billing/src/org/labkey/ehr_billing/EHR_BillingManager.java
+++ b/ehr_billing/src/org/labkey/ehr_billing/EHR_BillingManager.java
@@ -35,6 +35,8 @@ import org.labkey.api.query.InvalidKeyException;
 import org.labkey.api.query.QueryService;
 import org.labkey.api.query.QueryUpdateServiceException;
 import org.labkey.api.security.User;
+import org.labkey.api.data.AuditConfigurable;
+import org.labkey.api.gwt.client.AuditBehaviorType;
 
 import java.sql.SQLException;
 import java.util.ArrayList;
@@ -62,8 +64,15 @@ public class EHR_BillingManager
     public List<String> deleteBillingRuns(User user, Container container, Collection<String> pks, boolean testOnly) throws SQLException, QueryUpdateServiceException, BatchValidationException, InvalidKeyException
     {
         TableInfo invoice = QueryService.get().getUserSchema(user, container,EHR_BillingSchema.NAME).getTable(EHR_BillingSchema.TABLE_INVOICE);
+        if (invoice.supportsAuditTracking())
+            ((AuditConfigurable) invoice).setAuditBehavior(AuditBehaviorType.NONE);
+
         TableInfo invoiceRuns = QueryService.get().getUserSchema(user, container,EHR_BillingSchema.NAME).getTable(EHR_BillingSchema.TABLE_INVOICE_RUNS);
+
         TableInfo invoicedItems = QueryService.get().getUserSchema(user, container,EHR_BillingSchema.NAME).getTable(EHR_BillingSchema.TABLE_INVOICED_ITEMS);
+        if(invoicedItems.supportsAuditTracking())
+            ((AuditConfigurable)invoicedItems).setAuditBehavior(AuditBehaviorType.NONE);
+
         TableInfo miscCharges = EHR_BillingSchema.getInstance().getSchema().getTable(EHR_BillingSchema.TABLE_MISC_CHARGES);
 
         //create filters
--- a/ehr_billing/src/org/labkey/ehr_billing/EHR_BillingManager.java
+++ b/ehr_billing/src/org/labkey/ehr_billing/EHR_BillingManager.java
@@ -61,7 +61,7 @@ public class EHR_BillingManager
         return _instance;
     }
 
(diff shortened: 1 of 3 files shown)


ehr_billing/src/org/labkey/ehr_billing/EHR_BillingManager.java
```diff
@@ -52,43 +59,57 @@ public static EHR_BillingManager get()
...
             TableSelector tsMiscCharges2 = new TableSelector(miscCharges, miscChargesFilter, null);
-            ret.add(tsMiscCharges2.getRowCount() + " records from misc charges will be removed from the deleted invoice, which means they will be picked up by the next billing period.  They are not deleted.");
+            ret.add(tsMiscCharges2.getRowCount() + " invoice records from misc charges will be removed from the deleted invoice, which means they will be picked up by the next billing period.  They are not deleted.");
         }
         else
         {
             try (DbScope.Transaction transaction = ExperimentService.get().ensureTransaction())
             {
-                long deleted1 = Table.delete(invoicedItems, invoiceRunFilter);
+                Map<String, Object>[] invoicedItemsRows = tsInvItems.getMapArray();
```
labkey-jeckels: Suggest using getMapCollection() instead. I know there's plenty of other usages in the code for getMapArray(), but mixing generics and arrays tends to be slightly uglier

labkey-jeckels: Or better yet, can we avoid doing the select and stick with the filter approach instead?

ehr_billing/src/org/labkey/ehr_billing/EHR_BillingManager.java
```diff
@@ -52,43 +59,57 @@ public static EHR_BillingManager get()
...
+                deleteInvoicedRows(invoicedItems, invoicedItemsRows, user, container);
+
```
labkey-jeckels: As discussed, recommend just doing an audit entry for the whole run's deletion, don't need separate audit entries for every invoicedItem row

ehr_billing/src/org/labkey/ehr_billing/EHR_BillingManager.java
```diff
@@ -52,43 +59,57 @@ public static EHR_BillingManager get()
         return _instance;
     }
 
-    public List<String> deleteBillingRuns(User user, Collection<String> pks, boolean testOnly)
+    public List<String> deleteBillingRuns(User user, Container container, Collection<String> pks, boolean testOnly) throws SQLException, QueryUpdateServiceException, BatchValidationException, InvalidKeyException
```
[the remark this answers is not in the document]
labkey-bpatel: getUpdateService().deleteRows() in deleteBillingRuns() throws SQLException, and throwing RuntimeSQLException still requires either catching or throwing SQLException.

deleteRows(...) signature in QueryUpdateService.java: 
List<Map<String,Object>> deleteRows(User user, Container container, List<Map<String, Object>> keys, @Nullable Map<Enum, Object> configParameters, @Nullable Map<String, Object> extraScriptContext)
            **throws InvalidKeyException, BatchValidationException, QueryUpdateServiceException, SQLException**;

ehr_billing/src/org/labkey/ehr_billing/EHR_BillingManager.java
labkey-bpatel: Done.
I've set audit logging to none for invoicedItems and invoice tables.
Also, briefly looked to see if I could change the audit log message to include no. of rows deleted from invoicedItems and invoice - but didn't find anything that was easy to do without creating a custom audit event and provider, unless you have suggestions if we want to go that route.

labkey-bpatel: This is now updated to use Table.delete()

ehr_billing/src/org/labkey/ehr_billing/EHR_BillingManager.java
```diff
@@ -1,18 +1,3 @@
-/*
```
labkey-jeckels: Looks like the copyright is disappearing. Can you restore?

ehr_billing/src/org/labkey/ehr_billing/EHR_BillingManager.java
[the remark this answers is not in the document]
labkey-jeckels: OK. I'd like to remove SQLException from that throws list, but here I recommend catching it and rethrowing it as a RuntimeSQLException.

ehr_billing/src/org/labkey/ehr_billing/EHR_BillingManager.java
labkey-bpatel: Restored!

ehr_billing/src/org/labkey/ehr_billing/EHR_BillingManager.java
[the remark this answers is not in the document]
labkey-bpatel: done!

PR #18, opened 2019-08-21
Merge r. 64108-64192 from wnprc18.3 svn branch.

labkey-bpatel:
Add calculated column 'isActive' to aliases.


labkey-bpatel:
There is a separate commit in wnprc_billing module that references this customizer. I missed merging this file somehow when I did a merge from wnprc18.3 svn to ehrModules previously
Thanks!

PR #27, opened 2019-10-29
Add more comments, and some misc code cleanup

labkey-jeckels:


ehr_billing/api-src/org/labkey/api/ehr_billing/pipeline/BillingPipelineJobProcess.java
```diff
@@ -22,6 +22,10 @@
 import java.util.List;
 import java.util.Map;
 
+/**
+ * One section of an EHR billing implementation that generates the line items of a particular type of changes,
```
labkey-ankurj: *particular type of charges?

labkey-jeckels: Typo fixed.

PR #33, opened 2019-12-20
Make sure we don't try to create users that already exist

labkey-tchad:


labkey-tchad:
Verified on SNPRC: https://teamcity.labkey.org/viewLog.html?buildId=876639&buildTypeId=LabKey_Trunk_LabkeyPremiumTrunk_GitModules_SnprcEhrSqlserver
The failures there are from before my change to APIUserHelper

PR #41, opened 2020-02-21
Epic 6798 : List Domain Designer Conversion

labkey-bpatel:
- Updates related to allowing binding of DomainKind specific options/properties to a bean & generifying DomainKind.


labkey-jeckels (approved):
This is OK, but may not be necessary if we introduce the convenience subclass so that most other subclasses doesn't have to specify <JSONObject>

labkey-jeckels (changes requested):
Not needed anymore, I think

labkey-bpatel:
Yup, you're right, no longer needed. Closing PR & will delete the fb.

PR #46, opened 2020-04-07
Switch from direct ClientDependency references to Suppliers

labkey-jeckels:


ehr/api-src/org/labkey/api/ehr/dataentry/AbstractDataEntryForm.java
```diff
@@ -59,7 +61,7 @@
     private String _javascriptClass = "EHR.panel.DataEntryPanel";
     private String _storeCollectionClass = "EHR.data.StoreCollection";
     private List<FormSection> _sections;
-    private LinkedHashSet<ClientDependency> _clientDependencies = new LinkedHashSet<>();
+    private List<Supplier<ClientDependency>> _clientDependencies = new ArrayList<>();
```
bbimber, line 64: In general, it is easy for Collections of ClientDependency to get redundant.  I dont know about this specific instance, but often child classes feed dependencies into some parent, and there is a lot of repetition.  that's the reasoning usually having LinkedHashSet over ArrayList.  Is there a specific reason for all these switches?

labkey-adam: We hold the suppliers in a list. We resolve the suppliers into a LinkedHashSet to address the redundancy.

bbimber: OK.  I sent this comment to Josh but might be worth more seeing: it seems we've basically evolve two totally separate styles of ClientDependency.  We have the original fromXXX(), and now supplierFromXXX().  Can the former be made private or otherwise obscured?  Why have two disparate styles of use exposed to modules?  other than the fact that we actually have a lot of code currently using the fromXXXX() methods, should code actually be using that anymore?

bbimber: Given how much code is changing, could ClientDependency.fromPath() simply be modified to return Supplier<ClientDependency> (same for fromModule(), etc)?  It seems like virtually all the code around this is changing anyway.  There are a finite number of places that actually need to deal with a plain ClientDependency, and make those finite number of places should actually resolve the Supplier into the actual ClientDependency?

I realize some code can legitimately pass around short-lived ClientDependencies, but having two totally different schemes doesnt sit right.

labkey-adam: I don't like two ways either, but I didn't really want to take on the `fromPath()` methods this go around, given how many usages there are/were. But Josh dove in and eliminated a lot of them. I also want to be careful with performance... seems like a waste to construct one Supplier, dereference it, and immediately throw it away for every dependency on every page render. I probably shouldn't worry about that. Plan is to get the current PRs merged and then circle back to deprecate and hopefully eliminate the non-Supplier methods.

ehr/src/org/labkey/ehr/EHRServiceImpl.java
```diff
@@ -275,9 +290,10 @@ public void registerClientDependency(ClientDependency cd, Module owner)
         _clientDependencies.put(owner, list);
     }
 
-    public Set<ClientDependency> getRegisteredClientDependencies(Container c)
+    @Override
+    public Set<Supplier<ClientDependency>> getRegisteredClientDependencies(Container c)
```
labkey-adam: List?

ehr/src/org/labkey/ehr/EHRServiceImpl.java
```diff
@@ -275,9 +290,10 @@ public void registerClientDependency(ClientDependency cd, Module owner)
...
     {
-        Set<ClientDependency> set = new LinkedHashSet<>();
+        Set<Supplier<ClientDependency>> set = new LinkedHashSet<>();
```
labkey-adam: List?

ehr/src/org/labkey/ehr/buttons/CageBulkEditButton.java
```diff
@@ -32,9 +34,9 @@ public CageBulkEditButton(Module owner)
     {
         super(owner, "Bulk Edit Cage/Divider Types", "EHR.window.BulkEditCageTypeWindow.buttonHandler(dataRegionName);");
 
-        LinkedHashSet cds = new LinkedHashSet<ClientDependency>();
-        cds.add(ClientDependency.fromModuleName("ehr"));
-        cds.add(ClientDependency.fromPath("ehr/window/BulkEditCageTypeWindow.js"));
+        List<Supplier<ClientDependency>> cds = new ArrayList<>();
+        cds.add(ClientDependency.supplierFromModuleName("ehr"));
+        cds.add(ClientDependency.supplierFromPath("ehr/window/BulkEditCageTypeWindow.js"));
         setClientDependencies(cds);
```
labkey-adam: `setClientDependencies(List.of(...))` would be more succinct. Just a suggestion...

ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
```diff
@@ -1080,9 +1081,9 @@ private void configureMoreActionsBtn(AbstractTableInfo ti, ButtonBarConfig cfg)
                     btnNameMap.put(newButton.getText(), newButton);
                     menuItems.add(newButton);
 
-                    for (ClientDependency cd : fact.getClientDependencies(ti.getUserSchema().getContainer(), ti.getUserSchema().getUser()))
+                    for (Supplier<ClientDependency> cd : fact.getClientDependencies(ti.getUserSchema().getContainer(), ti.getUserSchema().getUser()))
```
labkey-adam: Resolve through `ClientDependency.getClientDependencySet()`?

ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
```diff
@@ -1014,9 +1015,9 @@ private void configureTbarBtns(AbstractTableInfo ti, ButtonBarConfig cfg)
             existingBtnNames.add(newButton.getText());
 
             existingBtns.add(newButton);
-            for (ClientDependency cd : fact.getClientDependencies(ti.getUserSchema().getContainer(), ti.getUserSchema().getUser()))
+            for (Supplier<ClientDependency> cd : fact.getClientDependencies(ti.getUserSchema().getContainer(), ti.getUserSchema().getUser()))
```
labkey-adam: Resolve through `ClientDependency.getClientDependencySet()`?

labkey-adam (approved):

ehr/api-src/org/labkey/api/ehr/dataentry/AbstractDataEntryForm.java
bbimber: OK, so you and Josh are pretty far down this road and probably you dont want to swap.  

However, might it be easier to separate ClientDependency into two classes?  ClientDependency would be as-is, except drop the cache and all methods that actually resolve resources.  It simply needs to hold the string cache-key path.  This could happen without changing anything across most of LK.  There could be a new ClientDependencyResource class, which has all the methods that actually resolve files.  ClientDependency could have getClientDependencyResource(), which actually creates that class.  Very limited callers would actually ever do this.  ClientDependencyResource would have the code would deal with the cache, actually resolving the right file(s), etc.?

labkey-jeckels: @bbimber I think that split is pretty close to where we are now, except that we're using Supplier<ClientDependency> instead of a named class.

Either way, I agree with the importance of steering developers to use the appropriate variant if we're not able to completely hide the resolved version.

ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
labkey-jeckels: Yes, thanks. That eliminates the concern about it resolving to null.

PR #65, opened 2020-09-18
Issue 41428: ONPRC_EHRTest.testNecropsyRequestFlow

labkey-jeckels:
Don't worry about the time component, which is working fine in practice for users. Just open the request for tomorrow.

### DIFF
--- a/ehr/test/src/org/labkey/test/tests/onprc_ehr/ONPRC_EHRTest.java
+++ b/ehr/test/src/org/labkey/test/tests/onprc_ehr/ONPRC_EHRTest.java
@@ -2163,7 +2163,7 @@ public class ONPRC_EHRTest extends AbstractGenericONPRC_EHRTest
     public void testNecropsyRequestFlow()
     {
         String animalId = "12345";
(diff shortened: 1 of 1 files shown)


bbimber:
The issue is that requests need to be made for a future date, not today?

labkey-jeckels:
Yes, that's the failure. There's a time-based component to this where the test passes at certain times of day but I don't think it's worth troubleshooting at that level of detail

labkey-sweta (approved):

bbimber:
@labkey-jeckels OK.  if I had to make a guess, it could be time zones and serialization into ISO date formats.  i spent a lot of time debugging team city tests back in the day around this. since JSON doesnt have a proper way to represent dates, they get serialized into strings and parsed on both ends.  i cant speak exactly to this failure, but it is often easy for an 8 hour shift between local time and GMT to result in the date getting bumped +/- 1 day.  i thought a lot of this had been fixed in EHR (there should be test cases around it).  not knowing any more than this thread, i agree it's maybe not worth debugging much. i would just be aware of that possibility.

PR #74, opened 2020-10-12
Merge from onprc19.1 & onprc19.1Prod r.63271 to 65949

labkey-bpatel:
#### Rationale
Merge changes from onprc19.1 r.63271 to 65757
Merge changes from onprc19.1Prod r. 65762 to 65949

#### Related Pull Requests
https://github.com/LabKey/onprcEHRModules/pull/32
https://github.com/LabKey/platform/pull/1626
https://github.com/LabKey/DiscvrLabKeyModules/pull/64
https://github.com/LabKey/LabDevKitModules/pull/55
https://github.com/LabKey/sampleManagement/pull/381
https://github.com/LabKey/biologics/pull/712
https://github.com/LabKey/labbook/pull/53
https://github.com/LabKey/wnprc-modules/pull/38


EHR_ComplianceDB/resources/schemas/dbscripts/sqlserver/EHR_ComplianceDB-12.39-12.40.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE ehr_compliancedb.Employees ALTER COLUMN Notes NVARCHAR (4000);
```
labkey-jeckels, line 1: We should do a matching Postgres script

ehr/resources/queries/study/treatment_order.query.xml
```diff
@@ -60,7 +60,7 @@
                         <columnTitle>Frequency</columnTitle>
                         <fk>
                             <fkDbSchema>ehr_lookups</fkDbSchema>
-                            <fkTable>treatment_frequency</fkTable>
+                            <fkTable>Treatment_Frequencies_Active</fkTable>
```
labkey-jeckels: This query looks to be defined via the schema browser for ONPRC. We will need to add it as a custom query in the EHR module to avoid breaking everyone else. Looks like no .query.xml is required, just the SQL:

Select * from ehr_lookups.treatment_Frequency
Where active = 'true'

ehr/resources/schemas/dbscripts/sqlserver/EHR_Lookups-17.20-17.21.sql
```diff
@@ -0,0 +1,3 @@
+/* File copied from EHR version upgrade to 17.21 in onprc17.2Prod branch */
+
+ALTER TABLE ehr_lookups.flag_categories ADD datedisabled DATETIME;
```
labkey-jeckels: This file isn't needed anymore. ONPRC is already on 19.1, and bootstrapping servers will run the rollup script.

ehr/resources/schemas/dbscripts/sqlserver/a_ehr_lookups-17.20-17.21.sql
```diff
@@ -0,0 +1,3 @@
+/* File copied from EHR version upgrade to 17.21 in onprc17.2Prod branch */
+
+ALTER TABLE ehr_lookups.flag_categories ADD datedisabled DATETIME;
```
labkey-jeckels: This file isn't needed anymore. ONPRC is already on 19.1, and bootstrapping servers will run the rollup script.

ehr/resources/schemas/dbscripts/sqlserver/ehr-17.21-17.30.sql
```diff
@@ -0,0 +1,40 @@
...
+ * Copyright (c) 2017 LabKey Corporation
+ *
+ * Licensed under the Apache License, Version 2.0: http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+/* Copy of file to deal with EHR version upgrade to 17.21 in onprc17.2Prod branch */
+
+/* ehr-17.20-17.21.sql */
+
+CREATE TABLE ehr.observation_types (
```
labkey-jeckels: This file isn't needed anymore. ONPRC is already on 19.1, and bootstrapping servers will run the rollup script.

ehr/resources/schemas/ehr_lookups.xml
```diff
@@ -2710,6 +2710,9 @@
                     <fkColumnName>entityid</fkColumnName>
                 </fk>
             </column>
+            <column columnName="datedisabled">
```
labkey-jeckels: This is a dupe from lines 2701-2703

ehr/resources/schemas/ehr_lookups.xml
```diff
@@ -2754,6 +2757,9 @@
...
```
labkey-jeckels: Looks like a bad merge. This column doesn't exist on this table.

ehr/resources/views/groupOverlaps.html
```diff
@@ -154,16 +154,19 @@
                 renderTo: this.qwpDiv.id,
                 schemaName: 'study',
                 queryName: queryName,
-                title: 'Group Members'
+                title: 'Group Members Test',
```
labkey-jeckels: This shouldn't include "Test"

ehr/src/org/labkey/ehr/EHRModule.java
```diff
@@ -256,6 +256,9 @@ public void moduleStartupComplete(ServletContext servletContext)
         EHRService.get().registerMoreActionsButton(new MarkCompletedButton(this, "study", "parentage", "End Selected Calls", true), "study", "parentage");
         EHRService.get().registerMoreActionsButton(new ExcelImportButton(this, "study", "parentage", "Import Data"), "study", "parentage");
 
+        EHRService.get().registerMoreActionsButton(new MarkCompletedButton(this, "study", "geneticAncestry", "End/Disable Selected", true), "study", "geneticAncestry");
```
labkey-jeckels: These should go into ONPRC_EHRModule instead.

ehr/src/org/labkey/ehr/EHRServiceImpl.java
```diff
@@ -456,7 +456,6 @@ public void registerOptionalClinicalHistoryResources(Module module)
         EHRService.get().registerLabworkType(new AntibioticSensitivityLabworkType(module));
         EHRService.get().registerLabworkType(new ChemistryLabworkType(module));
         EHRService.get().registerLabworkType(new HematologyLabworkType(module));
-        EHRService.get().registerLabworkType(new iStatLabworkType(module));
```
labkey-jeckels, line 459: To remove this here, we need to add it to WNPRC_EHRModule.doStartupAfterSpringConfig() so they don't lose it.

ehr/src/org/labkey/ehr/EHRServiceImpl.java
```diff
@@ -443,7 +443,7 @@ public void registerOptionalClinicalHistoryResources(Module module)
         EHRService.get().registerHistoryDataSource(new DefaultArrivalDataSource(module));
         EHRService.get().registerHistoryDataSource(new DefaultAssignmentEndDataSource(module));
         EHRService.get().registerHistoryDataSource(new DefaultBirthDataSource(module));
-        EHRService.get().registerHistoryDataSource(new DefaultClinicalRemarksDataSource(module));
+       // EHRService.get().registerHistoryDataSource(new DefaultClinicalRemarksDataSource(module));
```
labkey-jeckels: To remove this here, we need to add it to WNPRC_EHRModule.doStartupAfterSpringConfig() so they don't lose it.

Also, let's delete the line instead of commenting it out

labkey-jeckels (changes requested):
Overall code is looking pretty good.

EHR_ComplianceDB/resources/schemas/dbscripts/sqlserver/EHR_ComplianceDB-12.39-12.40.sql
labkey-bpatel: ah! yes, good catch! now done!

ehr/resources/queries/study/treatment_order.query.xml
labkey-bpatel: done

ehr/resources/schemas/dbscripts/sqlserver/EHR_Lookups-17.20-17.21.sql
labkey-bpatel: now removed

ehr/resources/schemas/dbscripts/sqlserver/a_ehr_lookups-17.20-17.21.sql
labkey-bpatel: now removed

ehr/resources/schemas/dbscripts/sqlserver/ehr-17.21-17.30.sql
labkey-bpatel: removed

ehr/resources/schemas/ehr_lookups.xml
labkey-bpatel: removed

ehr/resources/schemas/ehr_lookups.xml
labkey-bpatel: bad merge indeed! removed

ehr/resources/views/groupOverlaps.html
labkey-bpatel: done

ehr/src/org/labkey/ehr/EHRModule.java
labkey-bpatel: done

ehr/src/org/labkey/ehr/EHRServiceImpl.java
labkey-bpatel: done

ehr/src/org/labkey/ehr/EHRServiceImpl.java
labkey-bpatel: done

EHR_ComplianceDB/resources/schemas/dbscripts/postgresql/EHR_ComplianceDB-12.39-12.40.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE ehr_compliancedb.Employees ALTER COLUMN Notes TYPE NVARCHAR (4000);
```
labkey-jeckels: Postgres needs VARCHAR, not NVARCHAR

labkey-jeckels (changes requested):
One Postgres syntax issue, otherwise all set

ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
```diff
@@ -638,18 +615,36 @@ private void customizeStudyData(AbstractTableInfo ti)
...
+
+        // Hack the metadata for the length of the remark column, which has
+        // been altered in the DB so that it doesn't match with the standard
+        // "remark" property. See ONPRC ticket 33848 and EHRManager.ensureDatasetPropertyDescriptors()
+        TableInfo realTable = StorageProvisioner.getSchemaTableInfo(domain);
+        MutableColumnInfo remarkCol = (MutableColumnInfo) realTable.getColumn("remark");
+        remarkCol.setLocked(false);
+        remarkCol.setScale(1000000);
+        remarkCol.setLocked(true);
+    }
```
labkey-bpatel: @labkey-jeckels This is throwing below error during study import, since looks like remark column is "locked" and cannot be unlocked/modified this way. For ONPRC's use case, can we just add the scale for 'remark' column in datasets_metadata.xml?

java.lang.IllegalStateException: Can't unlock a ColumnInfo: remark
    at org.labkey.api.data.BaseColumnInfo.setLocked(BaseColumnInfo.java:2179)
    at org.labkey.ehr.table.DefaultEHRCustomizer.expandRemarksColumn(DefaultEHRCustomizer.java:630)
    at org.labkey.ehr.table.DefaultEHRCustomizer.customizeEncountersTable(DefaultEHRCustomizer.java:644)
    at org.labkey.ehr.table.DefaultEHRCustomizer.customizeDataset(DefaultEHRCustomizer.java:463)
    at org.labkey.ehr.table.DefaultEHRCustomizer.customize(DefaultEHRCustomizer.java:123)

EHR_ComplianceDB/resources/schemas/dbscripts/postgresql/EHR_ComplianceDB-12.39-12.40.sql
labkey-bpatel: done

ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
labkey-jeckels: In my testing we can just get rid of this completely. The length validation code now allows the length through based on the underlying column in the table.

ehr/resources/reports/schemas/study/Pedigree/Pedigree.r
```diff
@@ -13,10 +13,10 @@ library(Rlabkey)
 #str(labkey.data);
 #warnings();
 
-labkey.acceptSelfSignedCerts();
+labkey.setCurlOptions(ssl_verifypeer = FALSE, ssl_verifyhost = FALSE)
 
 
-if ((length(labkey.data$id) == 0) | (is.na(labkey.data$dam) & is.na(labkey.data$sire))){
+if ((length(labkey.data$id) == 0) || (is.na(labkey.data$dam) & is.na(labkey.data$sire))){
```
bbimber: the original was probably wrong in multi-animal cases, but I might write this like:

if ((length(labkey.data$id) == 0) || all(is.na(labkey.data$dam)) & is.na(labkey.data$sire)))){
 
which would require all rows to lack dam+sire.

ehr/resources/queries/study/treatment_order.query.xml
```diff
@@ -60,7 +60,7 @@
                         <columnTitle>Frequency</columnTitle>
                         <fk>
                             <fkDbSchema>ehr_lookups</fkDbSchema>
-                            <fkTable>treatment_frequency</fkTable>
```
bbimber, line 63: Comment: in general I dont like to encode the FK using a target that is filtered. This will be that legacy data now has broken FKs. In the situation where the desired behavior is to exclude inactive treatment_frequency records, I typically do one of:

1) in the data entry metadata, add this filter. this would mean that any EHR-related entry code would limit new records to the active frequencies only

2) in some cases you can add "<fkDisplayColumnName useRawValue="true"/>", which would at least mean the data region would render a reasonable value. i forget what the raw values of treatment_frequency are, but if they're strings maybe that's viable here.

ehr/resources/views/groupOverlaps.html
```diff
@@ -154,16 +154,19 @@
                 renderTo: this.qwpDiv.id,
                 schemaName: 'study',
                 queryName: queryName,
-                title: 'Group Members'
+                title: 'Group Members',
```
bbimber: why trailing comma

ehr/resources/views/groupOverlaps.html
```diff
@@ -154,16 +154,19 @@
...
             if (!start){
                 Ext4.Msg.alert('Error', 'Must provide a start date');
                 return;
             }
 
+            // By Kollil, 3/22/2018 : In 17.2 the filter is not working. Removed the the line 168 and added the filterArray at line 169.
+            // The Animal group search is working fine now after this change.
             if (groupName)
-                config.removeableFilters = [LABKEY.Filter.create('groupId/name', groupName, LABKEY.Filter.Types.EQUAL)];
+                //config.removeableFilters = [LABKEY.Filter.create('groupId/name', groupName, LABKEY.Filter.Types.EQUAL)];
```
bbimber, line 168: isnt userFilters the new name for removeableFilters?  this change prevents the user from removing the group name filter, doesnt it?

ehr/resources/web/ehr/window/CreateTaskFromIdsWindow.js
```diff
@@ -234,7 +234,7 @@ Ext4.define('EHR.window.CreateTaskFromIdsWindow', {
         var type = this.down('#type').getValue().type;
         if (type == 'createNew'){
             return {
-                date: this.down('#date').getValue(),
```
bbimber, line 237: dont you want to respect 'date'? otherwise is it going to default to today's date?

ehr/resources/web/ehr/window/CreateTaskFromRecordsWindow.js
```diff
@@ -346,7 +346,7 @@ Ext4.define('EHR.window.CreateTaskFromRecordsWindow', {
         EHR.Utils.createTask({
             initialQCState: 'Scheduled',
             existingRecords: existingRecords,
-            taskRecord: {date: date, assignedTo: assignedTo, category: 'task', title: title, formType: this.formType},
```
bbimber, line 349: same question as above. my knowledge of usage here is out of date, but the original scheme was to create the dataset records where 'date' is the date that work is expected to happen. did something change here?

bbimber (changes requested):

ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
labkey-bpatel: done

ehr/resources/reports/schemas/study/Pedigree/Pedigree.r
labkey-bpatel: Thanks for the updated snippet. Now updated!

ehr/resources/views/groupOverlaps.html
labkey-bpatel: now removed!

ehr/resources/queries/study/treatment_order.query.xml
labkey-bpatel: Good point, change is now reverted with an addition of a filter which will get applied only during insert and update, and not when viewing the data.

ehr/resources/views/groupOverlaps.html
labkey-bpatel: I believe that is the intention given that user enters filter criteria in a form, which then results in a filtered grid in a different panel, having removable filters on this resultant grid would be confusing.

ehr/resources/web/ehr/window/CreateTaskFromIdsWindow.js
labkey-bpatel: There is no 'date' column in ehr.tasks, there is a however a 'duedate' column, so this correction makes sense.

ehr/resources/web/ehr/window/CreateTaskFromRecordsWindow.js
labkey-bpatel: There is no 'date' column in ehr.tasks, there is a however a 'duedate' column, so this correction makes sense.

ehr/resources/views/groupOverlaps.html
bbimber: I think it's clear that if there was a bug preventing the filter from being applied, that's a problem.

I can see the argument on there being a filter panel above, but by that same logic then shouldnt start/end date also be immutable? It is possible for the top filter fields and DR to get out of sync.  

This is a small enough change that it's probably not a big deal, but in general I would tend toward preserving capabilities for the user. For example, what is they want to do another filter type?  More than one group?  Contains? etc.

ehr/resources/web/ehr/window/CreateTaskFromIdsWindow.js
bbimber: OK, i buy that. In reality I would bet that 'date' is often left alone (so defaulting to today on the client-side), and the server-side is going to default duedate to today. I dont think duedate ends up being used for a lot in practice. That all said, this fix is probably the right thing.

ehr/resources/web/ehr/window/CreateTaskFromRecordsWindow.js
bbimber: discussion above covers this

ehr/resources/queries/study/treatment_order.query.xml
bbimber: When did this LK feature get added to support <filters> in the query.xml?   That's useful....

labkey-bpatel: Looks like it got added early last year

ehr/test/src/org/labkey/test/tests/onprc_ehr/AbstractGenericONPRC_EHRTest.java
```diff
@@ -97,6 +97,13 @@ protected EHRClientAPIHelper getApiHelper()
         return new EHRClientAPIHelper(this, getContainerPath());
     }
 
+
+    @Override
+    protected boolean skipStudyImportQueryValidation()
+    {
+        return true;
```
labkey-jeckels: Is this still needed?

labkey-jeckels (approved):
One question but looks good.

ehr/test/src/org/labkey/test/tests/onprc_ehr/AbstractGenericONPRC_EHRTest.java
bbimber: Is this for test length / performance reasons?  Is at least one of the ONPRC EHR tests doing a query validation?

labkey-bpatel: This was added temporarily for query validation to pass so i could fix actual test failures but had missed to removed it after. Now removed, and query failures are fixed as well.

PR #77, opened 2020-10-19
Improve load time for data entry forms, from 10+ seconds to ~1 sec

labkey-jeckels:
#### Rationale
EHR data entry forms sometimes take a long time to load, even though the datasets have indices on the columns that are being used to filter the big tables (requestId, taskId). The problem is that SQL Server isn't choosing to use the index when the value is being passed as a JDBC parameter instead of a string literal in the SQL

#### Changes
* Add a redundant filter, which makes SQL Server choose a much faster execution plan

### DIFF
--- a/ehr/resources/web/ehr/panel/TaskDataEntryPanel.js
+++ b/ehr/resources/web/ehr/panel/TaskDataEntryPanel.js
@@ -20,6 +20,10 @@ Ext4.define('EHR.panel.TaskDataEntryPanel', {
         cfg = this.callParent(arguments);
         cfg.filterArray = cfg.filterArray || [];
         cfg.filterArray.push(LABKEY.Filter.create('taskid', this.taskId, LABKEY.Filter.Types.EQUALS));
+        // A redundant filter that can help the DB choose a more efficient query plan. SQLServer does better with this
+        // filter in choosing to use the index on the TaskId column, even if the value is being passed as a JDBC
+        // parameter instead of being embedded in the SQL as a string literal.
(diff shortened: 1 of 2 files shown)


ehr/resources/web/ehr/panel/TaskDataEntryPanel.js
```diff
@@ -20,7 +20,11 @@ Ext4.define('EHR.panel.TaskDataEntryPanel', {
         cfg = this.callParent(arguments);
         cfg.filterArray = cfg.filterArray || [];
         cfg.filterArray.push(LABKEY.Filter.create('taskid', this.taskId, LABKEY.Filter.Types.EQUALS));
+        // A redundant filter that can help the DB choose a more efficient query plan. SQLServer does better with this
+        // filter in choosing to use the index on the TaskId column, even if the value is being passed as a JDBC
+        // parameter instead of being embedded in the SQL as a string literal.
+        cfg.filterArray.push(LABKEY.Filter.create('taskid', this.requestId, LABKEY.Filter.Types.NOT_MISSING));
```
cnathe: The second param shouldn't matter for the NOT_MISSING filter type, but this one should like either be `this.taskId`, for consistency, or `null`.

cnathe (changes requested):

ehr/resources/web/ehr/panel/TaskDataEntryPanel.js
labkey-jeckels: Indeed. Copy/paste error now corrected

cnathe (approved):

PR #78, opened 2020-10-20
Don't allow duplicate account alias values

labkey-jeckels:
#### Rationale
Having duplicate account records will break many billing queries. No existing centers have duplicate values in use.

#### Related Pull Requests
* https://github.com/LabKey/tnprc_billing/pull/96

#### Changes
* Change from standard index to a unique constraint

### DIFF
--- a/ehr_billing/resources/schemas/dbscripts/postgresql/ehr_billing-20.001-20.002.sql
+++ b/ehr_billing/resources/schemas/dbscripts/postgresql/ehr_billing-20.001-20.002.sql
@@ -1,0 +1,4 @@
+-- Convert from a plain index to a unique constraint
+SELECT core.fn_dropifexists ('aliases', 'ehr_billing', 'INDEX', 'EHR_BILLING_ALIASES_INDEX');
+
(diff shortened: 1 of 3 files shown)


labkey-ankurj (approved):

bbimber:
i dont actually care all that much here, but why not make it the PK?  Wouldnt that be a lot more convenient for code?  See ContainerScopedTable, for example.

labkey-jeckels:
@bbimber I considered that and it's probably a good idea in the future. But I'm trying to make a marginally more targeted fix for 20.7, and I wasn't quite ready to enforce a server-wide uniqueness on the account number (I just did it at the container level)

labkey-jeckels:
@bbimber ContainerScopedTable and it's "virtual" PK may be a good way to handle this, which avoids the server-wide uniqueness aspect...

bbimber:
since EHRs dont use workbooks this also probably works. just in general i'm leary of databse-level FKs and labkey. that's my usage pattern though - i bet this probably works fine here.  it's been years, but i really thought i would have written some check into the aliases, since obvious dupes is a problem.  anyway, your concern these days, not mine.

PR #81, opened 2020-10-30
Fix ONPRC_BillingTest

labkey-jeckels:
#### Rationale
We are now creating the linked schema already in the base class's setup. Not sure if there was another failure that had caused the test to be muted, but this is the only fix needed.

#### Changes
* Remove second creation of the onprc_billing_public schema


labkey-bpatel (approved):
Makes sense, thanks for cleaning this up!

PR #89, opened 2020-11-13
Issue 83: Hook to push center specific properties

labkey-ankurj:
#### Rationale
This PR provides a hook for different centers to specify custom properties in trigger script java helper


ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -124,6 +124,7 @@
 
     //NOTE: consider moving these to SharedCache, to allow them to be shared across scripts, yet reset from admin console
     private Map<Integer, String> _cachedAccounts = new HashMap<>();
+    private static Map<String, String> _centerCustomProps;
```
labkey-jeckels: This needs some comments, both for its purpose and what properties are currently supported. I recommend initializing to an empty map and annotating at @NotNull so you don't have to null check when you use it.

labkey-jeckels (changes requested):

labkey-jeckels (approved):

PR #124, opened 2021-02-10
Bugfix to EHR Index Creation When Column Does not Exist

bbimber:
I may or may not be targeting the right branch here. 

EHR modules can register indexes they expect EHR to create on datasets. The intent of this code is to be tolerant and skip that index if the target columns dont exist, which could have wither different center-specific studies. The problem is that this check doesnt currently account for the 'include' columns. This PR extends the code to abort if the 'include' columns are not found.


labkey-jeckels:
Closing in preference of https://github.com/LabKey/ehrModules/pull/126 which has the right branch name

PR #126, opened 2021-02-12
Ticket 42449: Fix for index creation when a column doesn't exist

labkey-jeckels:
#### Rationale
The EHR code creates indices on certain columns which may not be present in some configurations

#### Changes
* Don't create the index if the column doesn't exist

### DIFF
--- a/ehr/src/org/labkey/ehr/EHRManager.java
+++ b/ehr/src/org/labkey/ehr/EHRManager.java
@@ -818,7 +818,6 @@ public class EHRManager
                     {
                         if (realTable.getColumn(col) == null)
                         {
-                            //messages.add("Dataset: " + d.getName() + " does not have column " + col + ", so indexing will be skipped");
                             missingCols = true;
                         }
                     }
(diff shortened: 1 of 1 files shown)


ehr/src/org/labkey/ehr/EHRManager.java
```diff
@@ -828,6 +823,18 @@ else if (d.getName().equalsIgnoreCase("blood"))
                         }
                     }
 
+                    if (includedCols != null)
+                    {
+                        for (String col : includedCols)
+                        {
+                            if (realTable.getColumn(col) == null)
+                            {
+                                //messages.add("Dataset: " + d.getName() + " does not have column " + col + ", so indexing will be skipped");
```
labkey-ankurj: do we need commented code?

labkey-ankurj (approved):

ehr/src/org/labkey/ehr/EHRManager.java
labkey-jeckels: I suppose we can say goodbye

PR #127, opened 2021-02-16
Add gitignore for dependencies.txt

bbimber:
LabKey seems to dislike dependencies.txt files getting checked in. Therefore it seems like a good idea to add them to .gitignore, so they're not sitting in the 'unversioned files' list. This does this for ehrModules, which creates a dependencies file for Viral_Load_Assay

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ ehr/.project
 ehr/resources/web/ehr/backup
 /hidra
 /hidra_uw_intake
+**/dependencies.txt


bbimber:
@labkey-jeckels what do you think about this minor change?

labkey-jeckels:
Yes, good idea. We've got similar config in some of our other repos but not here. I created a LabKey-side FB and PR, copying the variant we're using in other places

PR #128, opened 2021-02-17
Allow center-specific modules to override the default EHR blood volume validation

labkey-jeckels:
#### Rationale
Some centers want to opt out of the shared blood validation and use their own algorithms for allowable volume

#### Changes
* Register function via TriggerManager.registerHandlerForQuery() instead of using onUpsert()

### DIFF
--- a/ehr/resources/queries/study/blood.js
+++ b/ehr/resources/queries/study/blood.js
@@ -58,7 +58,7 @@ function onInit(event, helper){
     });
 }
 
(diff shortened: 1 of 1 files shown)


kollil:
The change has been validated in staging.

labkey-bpatel (approved):

kollil:
Thank you for the approval.

labkey-ankurj (approved):

PR #131, opened 2021-02-19
Ignore generated dependencies.txt files in any of the EHR modules

labkey-jeckels:
#### Rationale
dependencies.txt are auto-generated by the build and don't need to be committed to Git. As originally proposed in https://github.com/LabKey/ehrModules/pull/127

#### Changes
* Ignore dependencies.txt
* Hidra is long gone, no need to keep ignoring it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 ehr/.project
 ehr/resources/web/ehr/backup
-/hidra
-/hidra_uw_intake
+*/resources/credits/dependencies.txt


labkey-jeckels:
@labkey-ians do you think we should make this a part of more repos?

bbimber:
@labkey-jeckels I've been saying this for a while. personally i like checking in the dependencies.txt files so they become a bellweather for changes that require updates to jars.txt, but i think i'm a minority here.

labkey-susanh (approved):

PR #147, opened 2021-03-25
Fix ehr sql script version numbering to get in sync with 21.00x module version.

labkey-bpatel:
#### Rationale
Latest ehr script created was ehr-20.001-20.002.sql (it should have been named ehr-21.001-21.002.sql), but the module version got bumped from 21.001 to 21.002, rendering ehr-20.001-20.002.sql useless if the ehr module v. was already at 21.000.

#### Changes
*Add another script ehr-21.002-21.003.sql that's creating the same column and constraint as in ehr-20.001-20.002.sql, and bump the module v. to 21.003

### DIFF
--- a/ehr/resources/schemas/dbscripts/postgresql/ehr-21.002-21.003.sql
+++ b/ehr/resources/schemas/dbscripts/postgresql/ehr-21.002-21.003.sql
@@ -1,4 +1,9 @@
--- contents of ehr-20.001-20.002.sql
+-- same contents as ehr-20.001-20.002.sql.
+-- Rationale: When ehr-20.001-20.002.sql was created, the module version was bumped from 21.001 to 21.002
+-- rendering ehr-20.001-20.002.sql useless if the ehr module v. was already at 21.00x in the db.
+-- The script should have been numbered ehr-21.001-21.002.sql instead of ehr-20.001-20.002.sql.
+-- This script is an correction attempt to get in sync with the ehr module v. 21.00x so that this script runs
+-- and below col and constraint gets added as intended.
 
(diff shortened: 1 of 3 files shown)


ehr/resources/schemas/dbscripts/postgresql/ehr-21.002-21.003.sql
```diff
@@ -0,0 +1,38 @@
+-- contents of ehr-20.001-20.002.sql
```
labkey-adam: This comment by itself is a bit confusing... I would update it to explain (briefly) why we're doing this "again"

ehr/resources/schemas/dbscripts/sqlserver/ehr-21.002-21.003.sql
```diff
@@ -0,0 +1,8 @@
+-- contents of ehr-20.001-20.002.sql
```
labkey-adam: Update comment

labkey-adam (changes requested):

ehr/resources/schemas/dbscripts/postgresql/ehr-21.002-21.003.sql
labkey-bpatel: done

ehr/resources/schemas/dbscripts/sqlserver/ehr-21.002-21.003.sql
labkey-bpatel: done

labkey-adam (approved):

PR #154, opened 2021-04-02
Correct casing on table and schema names so that Postgres detects when they exist

labkey-jeckels:
#### Rationale
The upgrade script is intending to only add the ehr_billing.misccharges.chargegroup column when it's not already present. However, it's not matching correctly so it tries to double-add them and fails.

#### Changes
* Use lower case to match against PG metadata storage

### DIFF
--- a/ehr_billing/resources/schemas/dbscripts/postgresql/ehr_billing-20.000-20.001.sql
+++ b/ehr_billing/resources/schemas/dbscripts/postgresql/ehr_billing-20.000-20.001.sql
@@ -9,7 +9,7 @@ BEGIN
     IF NOT EXISTS (
             SELECT column_name
             FROM information_schema.columns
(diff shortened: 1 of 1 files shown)


labkey-bpatel (approved):
Curious , is this causing an issue lately? 
I dropped misccharges.chargegroup col, and ran addChargeGroupToMiscCharges() from pgAdmin (without casing change) and it ran without issues.

labkey-jeckels:
Yes, WNPRC hit this. They already had the column, but it didn't find it when running the metadata query so it tried to re-add it, which failed.

labkey-bpatel:
Huh, good to know! Thanks for fixing it.
I tried running that query again with col present, and ran into an issue. Strange that it checks correctly (?!) when the col is not present.

PR #158, opened 2021-04-08
Fix project entry field

labkey-ankurj:
#### Rationale
When an animal Id is entered into a data entry form grid row, the store gets populated with animal's project and setValue sets the value of project field with first value in the list which is correct, the subsequent addition of rows adds the projects in the store list and the value of next animal with different project is not first in the list and hence incorrect. This PR fixes the aforementioned problem and also fixes the problem of projects not getting populated from second row onwards with tabbing. 

#### Related Pull Requests
* https://github.com/LabKey/LabDevKitModules/pull/75

#### Changes
* do not insert more than one project in grid store
* populate the project field only when project is there

### DIFF
--- a/ehr/resources/web/ehr/form/field/ProjectEntryField.js
+++ b/ehr/resources/web/ehr/form/field/ProjectEntryField.js
@@ -128,7 +128,7 @@ Ext4.define('EHR.form.field.ProjectEntryField', {
                     var storeProjects = Ext4.Array.filter(store.collect('project'), function(proj) {
                         return LABKEY.Utils.isNumber(proj);
                     });
-                    if (storeProjects.length > 0) {
+                    if (storeProjects.length === 1) {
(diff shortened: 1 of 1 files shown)


ehr/resources/web/ehr/form/field/ProjectEntryField.js
```diff
@@ -344,11 +344,9 @@ Ext4.define('EHR.form.field.ProjectEntryField', {
         fromClient: true
       });
 
-      this.store.insert(0, newRec);
```
bbimber, line 347: Why dont you want to add this record to the store? If I read this right, you will end up setting the value but lack a record in the store to back it? Plus, isnt the point of creating the record here, as opposed to in UserEditableCombo, that the model has additional EHR-specific fields which get populated?

bbimber (changes requested):

PR #168, opened 2021-04-18
Prevent shared EHR properties from moving out of /Shared

labkey-jeckels:
#### Rationale
Fix long-standing bug where running the EHR's check for missing properties in datasets ends up reparenting shared property descriptors from /Shared into the EHR folder

#### Changes
* Re-parent the property descriptors to /Shared after adding them to the dataset domains


labkey-martyp (approved):
Glad you found a repro.  This fix looks good.  Looks like it only reparents the newly added PDs to shared container?

labkey-jeckels:
Yes, all the other properties that are dataset-specific should continue to live in the EHR folder.

PR #169, opened 2021-04-19
Issue 128 - Fix timeout issue in deaths trigger script

labkey-ankurj:
#### Rationale
Deaths trigger script times out during ETL due to the update of all the rows in demographics on deaths. This PR fixes the problem by making the update rows call outside the trigger script scope.

#### Related Pull Requests
*

### DIFF
--- a/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
+++ b/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
@@ -884,7 +884,7 @@ public class TriggerScriptHelper
                 {
                     throw new RuntimeException(e);
                 }
-            }, DbScope.CommitTaskOption.POSTCOMMIT);
+            }, DbScope.CommitTaskOption.PRECOMMIT);
             transaction.commit();
         }
     }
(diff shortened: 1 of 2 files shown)


ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -869,9 +870,23 @@ public void updateDemographicsRecord(List<Map<String, Object>> updatedRows) thro
...
+                try
+                {
+                    ti.getUpdateService().updateRows(getUser(), getContainer(), newRows, keyRows, null, getExtraContext());
+                    EHRDemographicsService.get().getAnimals(getContainer(), ids);
+                }
+                catch (InvalidKeyException | BatchValidationException | QueryUpdateServiceException | SQLException e)
+                {
+                    throw new RuntimeException(e);
+                }
+            }, DbScope.CommitTaskOption.POSTCOMMIT);
```
labkey-jeckels: This should be PRECOMMIT. It'll run inside the transaction itself, instead of in a separate transaction. It shouldn't impact the trigger script's timing.

labkey-jeckels (approved):
One change, but no need for additional review

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
labkey-ankurj: done

labkey-jeckels (approved):

PR #170, opened 2021-04-20
Check allowRequestsInDistantFuture setting before rejecting requests

labkey-jeckels:
#### Rationale
ScriptHelper.js has long had an allowRequestsInDistantFuture property but it's not being used in triggers.js when rejecting a request's date

#### Changes
* Skip the check completely when shouldAllowRequestsInDistantFuture() is true, otherwise check based on  defaultAllowedDaysForFutureRequest

### DIFF
--- a/ehr/resources/scripts/ehr/ScriptHelper.js
+++ b/ehr/resources/scripts/ehr/ScriptHelper.js
@@ -89,6 +89,7 @@ EHR.Server.ScriptHelper = function(extraContext, event, EHR){
         removeTimeFromDate: false,
         removeTimeFromEndDate: false,
         allowRequestsInPast: false,
+        // deprecated - use defaultAllowedDaysForFutureRequest to set an explicit date cutoff
         allowRequestsInDistantFuture: false,
         allowDeadIds: false,
         allowAnyId: false,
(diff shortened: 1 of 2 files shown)


labkey-ankurj (approved):

bbimber:
@labkey-jeckels I appreciate the existing code doesnt make sense (ignoring allowRequestsInDistantFuture), but note that defaultAllowedDaysForFutureRequest could also be set. would it make as much sense to:

1) rename defaultAllowedDaysForFutureRequest to simply allowedDaysForFutureRequest
2) let table set this value to something (perhaps -1 or null for zero limit), which would accomplish the same thing?

The rationale is to reduce parameters. It also favors set something as a time limit.

labkey-jeckels:
@bbimber I considered using defaultAllowedDaysForFutureRequest. The use case that's driving this change is setting this property via the extraContext JSON config in the Java form definition. WNPRC has code that references the allowRequestsInDistantFuture property (though I'm not convinced it's working as intended either). Since this is a 20.11 change I opted for the slightly lower impact tweak, especially since the allowDatesInDistantPast property is actively used as well. But it was a bit of a tossup and I've got mixed feelings about the two options.

bbimber:
@labkey-jeckels i dont feel particularly strongly. my comments were motivated by trying to move toward simpler configuration scheme.

bbimber:
sigh....i also see 'skipRequestInPastCheck'. 

a middle option that moves toward reducing the properties might be to put this logic into:

getDefaultAllowedDaysForFutureRequest: function(){
            return scriptOptions.allowRequestsInDistantFuture ? -1 : scriptOptions.defaultAllowedDaysForFutureRequest;
},

and then interpret the result downstream. it still respects allowRequestsInDistantFuture, but reduces the scope. you could @deprecate it, for example.

but i also really am not that invested in whatever is done here.

labkey-jeckels:
@bbimber that's a good idea and makes it easier to consolidate across different feature branches. I pushed an update along those lines

PR #171, opened 2021-04-24
Mute snprc crawler links in compliance test

labkey-martyp:
#### Rationale
Compliance training test isn't doing a full EHR setup so this link isn't getting fully setup.  This is only an issue because all the PRCs are included in this test distribution, so we can just ignore this link in the crawler.

#### Changes
* Add getUncrawlableActions() to compliance test


labkey-martyp:
@labkey-sweta Per my comments in TC, this link can be muted in the crawler.  I tested this on multiple test runs with no crawler errors.

labkey-sweta (approved):

PR #177, opened 2021-05-04
Update test to use "enddate" column name instead of bogus "end" field

labkey-jeckels:
#### Rationale
A trigger script recently got stricter about requiring an "enddate" value. That pointed out that the test has long been using the invalid "end" name, which was ignored

#### Related Pull Requests
* https://github.com/LabKey/onprcEHRModules/pull/149

#### Changes
* Validate that insert without an enddate value fails, and that one with it succeeds

### DIFF
--- a/ehr/test/src/org/labkey/test/tests/onprc_ehr/ONPRC_EHRTest.java
+++ b/ehr/test/src/org/labkey/test/tests/onprc_ehr/ONPRC_EHRTest.java
@@ -74,6 +74,7 @@ import java.util.Map;
 import java.util.Set;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
 
 @Category({CustomModules.class, EHR.class, ONPRC.class})
 @BaseWebDriverTest.ClassTimeout(minutes = 60)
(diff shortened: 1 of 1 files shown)


ehr/test/src/org/labkey/test/tests/onprc_ehr/ONPRC_EHRTest.java
```diff
@@ -195,15 +196,28 @@ public void testAssignmentApi() throws Exception
...
+            protocolCountsCommand.execute(getApiHelper().getConnection(), getContainerPath());
+            fail("Should have gotten an exception due to missing end date");
+        }
+        catch (CommandException e)
+        {
+            assertEquals("Wrong failure message", "ERROR: Must enter Start and End dates", e.getMessage());
+        }
+
+        // Then try again with the problem corrected
+        protocolCountsRow.put("end", prepareDate(new Date(), 370, 0));
```
labkey-ankurj: after reading the rationale I am confused if column name here should be "enddate" now?

labkey-jeckels: Yes, it should. I pushed and update and queued a test run:

https://teamcity.labkey.org/buildConfiguration/LabKey_2011Release_Premium_Ehr_OnprcEhrSqlserver?branch=fixAssignmentTest&mode=builds

labkey-ankurj (approved):

PR #185, opened 2021-05-25
Created a new column to track training categories that are not requir…

Ohsudev:
…ed as training requirements. This set is for the EHR Module version of additional code.

#### Changes
Created a new column to track training categories that are not required as training requirements. This set is for the EHR Module version of additional code.


Ohsudev:
Creating a new feature branch with the same name as the feature branch that resides in "onprcehrmodules".

PR #211, opened 2021-07-20
Created several Compliance Employee input screens.

Ohsudev:
#### Changes
Created several Compliance Employee input screens.


Ohsudev:
No longer needed.

PR #224, opened 2021-08-25
Attempt to fix test timing

bbimber:
@labkey-jeckels I dont think I have permission to create branches here (i'm getting a 403 from the command line and the github UI doesnt let me), or I would create this on the LK fork.


bbimber:
@labkey-jeckels Hi josh - sorry this isnt more streamlined, but I really dont think I have permission to create branches on the LK side of ehrModules. I thought perhaps at one point I did, but this might have been on onprcEHRModules. Could I either get this permission, or could you create a new branch based release21.3-SNAPSHOT, named 21.7_fb_xxxxx, which I will target my test changes into?

These are small changes to the VL selenium test that seem to have worked on similar tests to make them less flaky.

labkey-jeckels:
@bbimber sorry, I saw your note last week but I got pulled into something else before I could respond. You're right - you've got write access to onprcEhrModules but not ehrModules. I've added you as a collaborator with write access to this repo. Do you want to create the branch on the LK side? Otherwise let me know and I can copy it over.

PR #225, opened 2021-08-26
Add support for the modules to opt in to navigate to data entry forms when clicking edit links in animal history reporting

labkey-ankurj:
#### Rationale
This PR adds support for the modules to opt in to navigate to data entry forms when clicking edit links in animal history reporting through `EHRService.addModuleRequiringFormEditUI()`

#### Related Pull Requests
*

### DIFF
--- a/ehr/src/org/labkey/ehr/EHRController.java
+++ b/ehr/src/org/labkey/ehr/EHRController.java
@@ -332,6 +332,10 @@ public class EHRController extends SpringActionController
         @Override
         public ModelAndView getView(EHRQueryForm form, BindException errors)
         {
+
+            // if the modules can opt in - look into isUseLegagyExt3EditUI
+            // find task id and form type
+            // send it to edit form url
             ensureQueryExists(form);
 
             _form = form;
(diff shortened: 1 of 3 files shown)


ehr/src/org/labkey/ehr/EHRController.java
```diff
@@ -415,7 +416,15 @@ else if (isReactForm(form.getSchemaName(), form.getQueryName()))
                     }
                 }
 
-                DetailsURL updateUrl = DetailsURL.fromString(detailsStr);
+                DetailsURL updateUrl;
+                if (EHRServiceImpl.get().isUseFormEditUI(getContainer()) && null != ti.getColumn("taskid"))
```
labkey-ankurj, line 420: @labkey-jeckels I think this will be useless when there is no value for task ids.

labkey-jeckels: Yes, agreed. Checking for the presence of that column is a good idea.

labkey-jeckels (approved):
See TNPRC PR for a rename suggestion, but otherwise looks good.

PR #227, opened 2021-08-30
Attempt to fix test timing (#224)

bbimber:
@labkey-jeckels thanks - this will make things much easier going forward


labkey-jeckels (changes requested):
It looks like clickAndGetExampleData() doesn't exist in 21.3.

PR #247, opened 2021-10-05
Added two new input screens and added one new tables.

Ohsudev:
#### Changes
EHR version of Compliance input screens.  Redesigned from previous versions of featue branch.

### DIFF
--- a/EHR_ComplianceDB/resources/schemas/dbscripts/postgresql/EHR_ComplianceDB-12.40-12.41.sql
+++ b/EHR_ComplianceDB/resources/schemas/dbscripts/postgresql/EHR_ComplianceDB-12.40-12.41.sql
@@ -1,16 +1,13 @@
 
-ALTER TABLE ehr_compliancedb.RequirementsPerCategory add trackingflag varchar(100);
-GO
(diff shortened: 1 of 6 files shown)


kollil (approved):
Is update on postgresql intentional?EHR_ComplianceDB/resources/schemas/dbscripts/postgresql/EHR_ComplianceDB-12.40-12.41.sql

labkey-jeckels:
@kollil Yes, we need to make sure that the SQLServer and Postgres versions of the schema stay consistent. Thanks for putting in the Postgres side, Raymond. However, there are differences between the DBs (Postgres doesn't have NVARCHAR, for example) that will need to be fixed. I assume you don't have Postgres setup, @Ohsudev, so I can take a crack at this later today.

labkey-jeckels:
@Ohsudev I pushed additional changes for Postgres compatibility (it uses SERIAL instead of IDENTITY and doesn't like GO statements). Let's make sure they're happy on TeamCity test runs but I hope that should be all that's needed.

labkey-jeckels (approved):

Ohsudev:
No longer needed.

PR #250, opened 2021-10-07
Clinical remarks data source

labkey-martyp:
#### Rationale
Clinical remarks categories value can be an integer so try to use title if it's available, then fall back to value.  Also refresh EHR data entry cache when EHR lookup sets are updated.

#### Related Pull Requests
* https://github.com/LabKey/nircEHRModules/pull/47

#### Changes
* Add change to DefaultClinicalRemarksDataSource.getCategoryText
* Add lookupsets table info and QUS with cache refresh

### DIFF
--- a/ehr/src/org/labkey/ehr/query/LookupSetsTable.java
+++ b/ehr/src/org/labkey/ehr/query/LookupSetsTable.java
@@ -5,6 +5,7 @@ import org.labkey.api.data.ContainerFilter;
 import org.labkey.api.data.TableInfo;
 import org.labkey.api.ldk.table.ContainerScopedTable;
 import org.labkey.api.query.BatchValidationException;
+import org.labkey.api.query.InvalidKeyException;
 import org.labkey.api.query.QueryUpdateService;
 import org.labkey.api.query.QueryUpdateServiceException;
 import org.labkey.api.query.SimpleUserSchema;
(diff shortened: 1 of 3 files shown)


ankurjuneja (approved):

labkey-jeckels (approved):

ehr/src/org/labkey/ehr/query/LookupSetsTable.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private class UpdateService extends ContainerScopedTable<SchemaType>.UpdateService
+    {
+        public UpdateService(SimpleUserSchema.SimpleTable<SchemaType> ti)
+        {
+            super(ti);
+        }
+
+        @Override
+        protected void afterInsertUpdate(int count, BatchValidationException errors)
```
labkey-jeckels, line 44: For completeness should hook into deleteRows() as well. truncateRows() gets called when the whole table is being cleared, but won't be invoked for individual rows being deleted.

labkey-jeckels (changes requested):

ehr/src/org/labkey/ehr/query/LookupSetsTable.java
labkey-martyp: Good call, added.

labkey-jeckels (approved):

PR #256, opened 2021-10-19
log non-date and non-string type date values

ankurjuneja:
#### Rationale
It is hard to determine the value of date that does not get normalized during ETLs. This change logs such values which are of type other than string and date.


ehr/resources/scripts/ehr/utils.js
```diff
@@ -180,18 +180,21 @@ EHR.Server.Utils = new function(){
...
             }
-            else if (!isNaN(val)){
-                // NOTE: i'm not sure if we should really attempt this.  this should really never happen,
-                // and it's probably an error if it does
-                normalizedVal = new Date(val);
-            }
             else {
-                if (val['getTime']){
-                    normalizedVal = new Date(val.getTime());
+                console.log('EHR trigger script is being passed a date with value: ' + val);
```
labkey-jeckels, line 184: This will log to the console (an improvement over the current behavior) and then fail the job with the same unhelpful error message.

I've asked Matthew to share the full log file. I checked the recording but he didn't scroll down quite far enough to show the root exception during the screensharing yesterday.

labkey-jeckels:
Does the platform PR effectively make this logging redundant?

ehr/resources/scripts/ehr/utils.js
ankurjuneja: This will give us the value of the date which we were eyeballing in granite and couldn't find.

ankurjuneja:
I think this PR will be useful for looking into the failures of locations ETL, the platform PR will log the error messages for arrivals ETL.

labkey-jeckels (approved):

PR #264, opened 2021-11-16
Issue 44283: Age is not calculated for new animals on snapshot

labkey-jeckels:
#### Rationale
The EHR makes heavy use of queries to calculate values like current animal age and weight via SQL queries wired up as lookups through the Id field. Those lookups only resolve correctly when the study module has already inserted a row into study.participant for the new animal/participant, so we shouldn't trust calculated demographic info until the study module has done its bookkeeping.

#### Changes
* When we know we're inserting a new demographics record, likely the first time we've seen an animal, refresh the demographics call an extra time.

### DIFF
--- a/ehr/resources/queries/study/Demographics.js
+++ b/ehr/resources/queries/study/Demographics.js
@@ -14,17 +14,17 @@ function onInit(event, helper){
     });
 }
 
-function onUpsert(helper, scriptErrors, row, oldRow){
+function onInsert(helper, scriptErrors, row) {
+    // If we're doing inserts in demographics (where we don't have an old row), don't fire the normal notifications
+    // of changes to refresh the demographics cache, as we have to do a special clearing after study has done its
+    // bookkeeping. See ticket 44283 and the clearing in TriggerScriptHelper.createDemographicsRecord()
+    helper.getExtraContext().skipAnnounceChangedParticipants = true;
+}
+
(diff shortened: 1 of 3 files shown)


ankurjuneja (approved):

bbimber:
@labkey-jeckels so this is being a little picky, but given that caching can be a fairly expensive process, could the trigger force that to happen during another point of the trigger life-cycle, like complete()? Does study.participant have triggers of its own?

labkey-jeckels:
Thanks for the review, Ben.

The current recaching is already happening in the complete() trigger script code, but the order of operations is such that the study bookkeeping happens after that, unfortunately.

I settled on this approach because it only does the extra cache clear on a brand new demographics record (not for updates of an existing demographics row), and because it queues the reload to happen asynchronously. I don't love it, but I couldn't find anything that wasn't a giant hack. I'll give it another look tomorrow morning with a fresh set of eyes though.

labkey-jeckels:
And I should add that the study code is doing a direct INSERT statement against the DB to create the record - see VisitManager.updateParticipants() if you're curious, so it's pretty far removed from supporting its own trigger.

labkey-jeckels:
See my followup commit - I found a reasonable way to avoid firing the normal cache clearing when inserting the new demographics row, which means that we avoid the extra reload

PR #269, opened 2021-11-24
Reload module-based TSVs as part of EHR module schema upgrades

labkey-jeckels:
#### Rationale
Make it easier to do dev-ops style loading of data when TSVs for simple lookups change in the module code

#### Related Pull Requests
* https://github.com/LabKey/nircEHRModules/pull/66

#### Changes
* New SQL script "stored proc" to reload TSV data: SELECT core.executeJavaUpgradeCode('importFromTsv;ehr;reports;/data/reports.tsv');
* Avoid creating multiple instances of upgrade code to declutter server bootstrap log files
* Ensure to reset transform state if truncating ETL


labkey-martyp (approved):
This change looks good. Nice feature.

@labkey-jeckels Not related to this PR but looking at this code for the first time, looks like we're not handling the case where ETL truncate fails.  We should probably log the error and not continue to the ETL in that case, don't you think?

labkey-jeckels:
@labkey-martyp Sure, sounds good. I improved the method signature (see two other PRs) instead of fishing out an undocumented property in the map. I find the error handling inside of the truncate method questionable (in some cases it captures warnings but continues to attempt truncation of other tables, and the truncation itself isn't transacted so who knows what state the tables will be if there's a problem) but I've left that for a future set of cleanup.

PR #271, opened 2021-11-25
Remove extra verbose logging added recently for troubleshooting

labkey-jeckels:
#### Rationale
The date logging is overwhelming server log. We've already found the problematic data and also added more centralized logging of the full problematic row.

#### Changes
* Back out the temporary logging


ankurjuneja (approved):
thanks! that was on my list.

PR #276, opened 2021-12-05
Prevent duplicate animals in pedigree plot

labkey-martyp:
#### Rationale
Pedigree plot is producing duplicate animals due to "unique" rows being thrown off by NA vs blank values in the display column in source data.  Need to change NA to blank.

#### Changes
* When creating the pedigree data frame, set NA display values to blank


labkey-jeckels (approved):
Thanks for tracking this one down.

PR #298, opened 2021-12-21
Speed up populating ehr_lookups.calendar from ~1m7s to ~3s

labkey-jeckels:
#### Rationale
Table.insert() isn't a fast way to populate ~30k rows

#### Changes
* Use a PreparedStatement and transaction to speed up populating ehr_lookups.calendar about 20x

### DIFF
--- a/ehr/src/org/labkey/ehr/EHRUpgradeCode.java
+++ b/ehr/src/org/labkey/ehr/EHRUpgradeCode.java
@@ -15,15 +15,16 @@
  */
 package org.labkey.ehr;
 
-import org.labkey.api.data.Table;
+import org.labkey.api.data.DbScope;
 import org.labkey.api.data.TableInfo;
 import org.labkey.api.data.UpgradeCode;
 import org.labkey.api.module.ModuleContext;
 
+import java.sql.Date;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
 import java.util.Calendar;
 import java.util.GregorianCalendar;
-import java.util.HashMap;
-import java.util.Map;
 
 public class EHRUpgradeCode implements UpgradeCode
 {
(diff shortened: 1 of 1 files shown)


ehr/src/org/labkey/ehr/EHRUpgradeCode.java
```diff
@@ -15,50 +15,53 @@
...
-            row.put("TargetDateTime", cal.getTime());
-            row.put("TargetDate", cal.getTime());
-            row.put("Year", cal.get(Calendar.YEAR));
-            // java.util.Calendar months are 0-based
-            row.put("Month", cal.get(Calendar.MONTH) + 1);
-            row.put("Day", cal.get(Calendar.DAY_OF_MONTH));
+            while (cal.get(Calendar.YEAR) < 2030)
+            {
+                stmt.setDate(1, new Date(cal.getTime().getTime()));
+                stmt.setDate(2, new Date(cal.getTime().getTime()));
```
labkey-martyp: You could use cal.getTimeInMillis here

labkey-martyp (approved):
Pretty great improvement. Looks useful in other contexts too

ehr/src/org/labkey/ehr/EHRUpgradeCode.java
labkey-jeckels: Yes, that's a little cleaner. Thanks.

labkey-matthewb (commented):
Yes!  Also just a reminder that ParameterMapStatement is also a very useful class.  Not better in this context, but still useful.

bbimber:
@labkey-jeckels, just curious (and this is not worth any effort if you dont already know anything about this) - this change includes both executing the work as a transaction and the switch from Table.insert() to PreparedStatement. In theory, the transaction part could have been implemented with Table.insert(), right? Any sense on how much of the benefit comes from using a transaction versus PreparedStatement? I'm curious b/c I agree this pattern might be useful elsewhere.

labkey-matthewb:
(sorry for posting to an old PR, I forgot that I had a mail search on and found this at the top of my inbox)

Table.insert() is very convenient for inserting one row.  It's a disaster for bulk insert.  a) each insert does a ton of metadata inspection to create a new SQL statement.  That work is not saved and is repeated for each row.  b) it inserts one at a time.  This is a disaster of round-trip and parsing overhead communicating with the SQL database.

bbimber:
@labkey-matthewb, very good to know. I can definitely think about a number of cases that use it in a loop like this. I'll make a note to check out a few of those...

labkey-jeckels:
@bbimber I don't have the benchmarking numbers for the individual pieces of this change anymore, but I think almost all of the boost was from migrating away from Table.insert() like Matt said. And yes, those two parts could be done independently.

bbimber:
ok, thanks

PR #300, opened 2021-12-23
Clean up intermittent test failures when requestid is null when form is submitted

labkey-jeckels:
#### Rationale
https://teamcity.labkey.org/buildConfiguration/LabKey_217Release_Premium_Ehr_OnprcEhrSqlserver/1649642?buildTab=artifacts#%2FONPRC_EHRTest2.tar.gz

Can't repro locally. Intent of code seems to be to log true mismatches instead of a race condition where it hasn't been set.

#### Changes
* Tolerate requestid being null without warning message on server


labkey-martyp (approved):

bbimber:
@labkey-jeckels I think your change is fine. This check was probably introduced years ago during development. I'm still surprised it's possible to have this condition since I thought requestId would be set when the client model was created.

labkey-jeckels:
@bbimber yes, it's always set early whenever I've tried to repro this.

PR #309, opened 2022-01-11
Attempt to improve ViralLoadAssayTest timing

bbimber:


bbimber:
An example of the intermittent failure this is trying to fix is here:

https://teamcity.labkey.org/buildConfiguration/LabKey_2111Release_External_Discvr_ExternalModulesTestSqlserver/1667969?showRootCauses=false&expandBuildChangesSection=true&expandBuildTestsSection=true

labkey-tchad (approved):
Looks reasonable.
It's not necessary but, if you pushed your branch to the LabKey fork, TeamCity would be able to verify the fix.

PR #325, opened 2022-02-08
Reduce memory use and improve perf when populating EHR demographics cache during ETL

labkey-jeckels:
#### Rationale
We had a production server run out of memory on a 4GB heap during a large ETL

#### Changes
* Copy list of animal IDs to avoid holding JS "native" array in memory, which can drag the whole JS trigger script scope through
* Use a DefaultSchema during demographics cache refresh so that schemas and tables can be cached


labkey-martyp (approved):
Nice, looks like improvements we could use in more places in EHR

PR #326, opened 2022-02-09
Restrict data entry forms for registered groups

ankurjuneja:
#### Rationale
This PR provides a way for modules to register a way to restrict a user/group to a set of data entry forms. 

#### Related Pull Requests
* https://github.com/LabKey/tnprc_ehr/pull/732

### DIFF
--- a/ehr/src/org/labkey/ehr/EHRController.java
+++ b/ehr/src/org/labkey/ehr/EHRController.java
@@ -61,7 +61,9 @@ import org.labkey.api.query.UserSchema;
 import org.labkey.api.query.ValidationError;
 import org.labkey.api.query.ValidationException;
 import org.labkey.api.resource.FileResource;
+import org.labkey.api.security.Group;
 import org.labkey.api.security.RequiresPermission;
+import org.labkey.api.security.SecurityManager;
 import org.labkey.api.security.permissions.AdminPermission;
 import org.labkey.api.security.permissions.DeletePermission;
 import org.labkey.api.security.permissions.ReadPermission;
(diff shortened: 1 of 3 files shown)


ehr/src/org/labkey/ehr/EHRController.java
```diff
@@ -141,9 +143,46 @@ public ApiResponse execute(GetDataEntryItemsForm form, BindException errors)
 
             Collection<DataEntryForm> forms = DataEntryManager.get().getForms(getContainer(), getUser());
             List<JSONObject> formJson = new ArrayList<>();
-            for (DataEntryForm def : forms)
+
+            // restrict forms if the user is in the restricted group
+            var userGroups = SecurityManager.getGroups(getContainer(), getUser());
```
labkey-jeckels, line 148: I was hoping that this code wouldn't know about the group names. Instead, it would just be asking a filter if a given form is acceptable for the current user.

ehr/src/org/labkey/ehr/EHRController.java
```diff
@@ -141,9 +143,46 @@ public ApiResponse execute(GetDataEntryItemsForm form, BindException errors)
...
+                    }
+                }
+            }
+
+            if (isUserInRestrictedGroup)
             {
-                formJson.add(def.toJSON(form.isIncludeFormElements()));
+                for (DataEntryForm f : forms)
+                {
+                    permittedForms.get(restrictedGroupName).forEach(permForm -> {
```
labkey-jeckels, line 171: I suggest matching on the getName() for the form instead of its class. It's possible that a single class could be registered with different configurations that produce different names

ehr/src/org/labkey/ehr/EHRServiceImpl.java
```diff
@@ -1011,4 +1027,25 @@ public void appendSNOMEDCols(AbstractTableInfo ti, String displayColumnName, Str
             ti.addColumn(simpleRawCol);
         }
     }
+
+    @Override
+    public void registerDataEntryFormAccess(Map<Module, Map<String, Set<Class<? extends DataEntryForm>>>> permittedForms)
```
labkey-jeckels, line 1032: This overall pattern looks good, but I propose that we create a DataEntryFormFilter interface with a single method, isAvailable(DataEntryForm, Container, User). TNPRC can implement that interface with its group/form name logic

labkey-jeckels (changes requested):

bbimber:
Isnt there already a mechanism to do this? unless i misunderstand, doesnt onprcEHR contain a lot of examples of form-specific testing, including on groups?

labkey-jeckels:
@bbimber you're right. Thanks. @ankurjuneja, we were so caught up in the permissions questions that I think we missed an easier way to handle this. For the forms that need to be hidden, I think you can override isAvailable(). If the super class thinks it's available, then also check if the user is in the group (available via _ctx.getUser()) and return false if they shouldn't see the form. It's a slightly different approach than most of the other places that are checking for extra permissions needed for that form, but it should work regardless.

bbimber:
yeah, that is basically what i was thinking. The FlagForm in ONPRC is probably among the more custom. I think there is an isVisible(), in case you want to just obscure it but let users open it is they have direct links.

in general, if you want to support some quirky, weird permission thing, ONPRC probably already blazed your trail...

PR #346, opened 2022-03-29
Cache only immutable collections

labkey-jeckels:
#### Rationale
New warnings in our caching layer pointed out the EHR code is caching mutable maps and sets. It's both thrilling and dangerous.

#### Changes
* Wrap collections in an unmodifiable layer before caching them

### DIFF
--- a/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
+++ b/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
@@ -516,6 +516,7 @@ public class TriggerScriptHelper
             }
         }
 
+        ret = Collections.unmodifiableMap(ret);
         DataEntryManager.get().getCache().put(cacheKey, ret);
 
         return ret.get(project);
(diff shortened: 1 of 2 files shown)


ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -516,6 +516,7 @@ public String getProtocolForProject(Integer project)
             }
```
labkey-adam, line 521: I don't think `ret.put()` is going to work very well once `ret` is unmodifiable. I think you'll need a similar pattern to the below.

labkey-adam (changes requested):
See comment on TriggerScriptHelper... I think new code will attempt to change an unmodifiable map.

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
labkey-jeckels: Fixed. Still a clunky approach to caching, but an incremental improvement.

labkey-adam (approved):

PR #355, opened 2022-04-07
Ticket 45145: CreatedBy fk overrides

labkey-martyp:
#### Rationale
Built in CreatedBy column fks overridden.  Causes users removed from the project not to resolve in the lookup.  Not sure why this was done initially.  I don't see anything related to data entry or reporting where this is necessary.

#### Changes
* Remove fk overrides

### DIFF
--- a/ehr/resources/queries/ehr/Tasks_DataEntry.query.xml
+++ b/ehr/resources/queries/ehr/Tasks_DataEntry.query.xml
@@ -45,11 +45,6 @@
                     </column>
                     <column columnName="createdby">
                         <columnTitle>Created By</columnTitle>
-                        <fk>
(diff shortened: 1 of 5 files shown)


bbimber:
if my memory is right, back whenever this was written LabKey was not as consistent about carrying metadata and lookups through queries. there are likely a lot of places in original EHR code where lookups are explicitly added in the query XML that are probably not strictly needed anymore.

labkey-martyp:
> if my memory is right, back whenever this was written LabKey was not as consistent about carrying metadata and lookups through queries. there are likely a lot of places in original EHR code where lookups are explicitly added in the query XML that are probably not strictly needed anymore.

Ok sounds good, thanks for the context Ben

labkey-jeckels (approved):

PR #356, opened 2022-04-08
Ticket 45056: Pedigree plot failing on a specific animal

labkey-martyp:
#### Rationale
Element-wise operator in if statement causing "argument is of length zero" errors for empty records.

#### Changes
* Change to logical operators

### DIFF
--- a/ehr/resources/reports/schemas/study/Pedigree/Pedigree.r
+++ b/ehr/resources/reports/schemas/study/Pedigree/Pedigree.r
@@ -15,7 +15,7 @@ library(Rlabkey)
 
 labkey.setCurlOptions(ssl_verifypeer = FALSE, ssl_verifyhost = FALSE)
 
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):

bbimber:
@labkey-jeckels are you sure here?

bbimber:
OK, so perhaps this is always executed with one ID; however, I might do this instead:

if ((nrow(labkey.data) == 0) | (all(is.na(labkey.data$dam)) & all(is.na(labkey.data$sire)))){

labkey-martyp:
> OK, so perhaps this is always executed with one ID; however, I might do this instead:
> 
> if ((nrow(labkey.data) == 0) | (all(is.na(labkey.data$dam)) & all(is.na(labkey.data$sire)))){

Thanks Ben that's a good recommendation.  I updated to that change for multiple Ids.

PR #370, opened 2022-04-29
21.11 fb purchasing line item numbers

gottfred:
#### Rationale
A new column is needed for giving each item in a purchase order a line item number

#### Related Pull Requests
The code added in the wnprc-modules relies on this new column
https://github.com/WNPRC-EHR-Services/wnprc-modules/pull/285

#### Changes
Added a new column to ehr_purchasing.lineItems and also added it to the schema xml file


labkey-jeckels (changes requested):
If you make the upgrade script 22.000-22.001 and set the version to 22.001 in the module code, we'll be able to merge it forward to 22.3 and beyond as-is. Otherwise, we'll have to do special handling because the schema version has already been bumped to 22.000 in newer branches.

gottfred:
> If you make the upgrade script 22.000-22.001 and set the version to 22.001 in the module code, we'll be able to merge it forward to 22.3 and beyond as-is. Otherwise, we'll have to do special handling because the schema version has already been bumped to 22.000 in newer branches.

I've made the requested changes

labkey-jeckels (approved):

PR #371, opened 2022-04-29
Update WNPRC pedigree plots for new visPedigree library

dnicolalde:
#### Rationale
Modifying use of RLabKey inside Pedigree plots for WNPRC

#### Related Pull Requests
* 

#### Changes
- Changing ssl.verifyhost to ssl_verifyhost to work with RLabKey 2.8.2


labkey-martyp (approved):
Looks good.  This is actually required due to a change in curl

PR #378, opened 2022-05-06
Option to do query validation at end of tests

labkey-martyp:
#### Rationale
Currently EHR tests don't do query validation in test postamble. This provides option to do that validation there.

#### Related Pull Requests
* https://github.com/LabKey/onprcEHRModules/pull/426

#### Changes
* Add shouldValidateQueries()


labkey-jeckels (approved):
Tests are failing again, so cleared to commit (an atypical comment in a PR review).

@labkey-tchad will the TeamCity config that enables the post-test validation automatically be part of future ONPRC test suite config so we don't lose it? Or should we revert the code that's running it at the end once the queries are fixed up?

PR #383, opened 2022-05-12
Viral assay load decimal precision

labkey-martyp:
#### Rationale
ExtJS defaults number inputs to two decimal places. Currently this is not configurable for the viral load assays. This PR sets the decimal precision to match the formatString (if defined) for extjs decimal columns in viral load assay results. Default is still two. Setting formatString on column metadata will set the number of decimal places.

#### Changes
* Default metadata for the assay results adds the decimalPrecision to any decimal or double values with a formatString defined

### DIFF
--- a/Viral_Load_Assay/src/org/labkey/viral_load_assay/Viral_Load_Manager.java
+++ b/Viral_Load_Assay/src/org/labkey/viral_load_assay/Viral_Load_Manager.java
@@ -16,6 +16,12 @@
 package org.labkey.viral_load_assay;
 
 import org.json.JSONObject;
+import org.labkey.api.data.ColumnInfo;
+import org.labkey.api.data.JdbcType;
+import org.labkey.api.data.TableInfo;
+import org.labkey.api.query.QueryService;
+import org.labkey.api.query.SchemaKey;
+import org.labkey.api.view.ViewContext;
 
 /**
  * Created with IntelliJ IDEA.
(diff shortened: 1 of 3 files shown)


bbimber:
@labkey-martyp This is reasonable, and I agree in theory it might be useful to have logic like this more centrally. However, one question: could code interpret the java format string and conditionalize the decimal places based on this? If this behavior were pushed more centrally, interpreting the column-specific format is cleaner than guessing a number and applying it everywhere.

labkey-martyp:
> @labkey-martyp This is reasonable, and I agree in theory it might be useful to have logic like this more centrally. However, one question: could code interpret the java format string and conditionalize the decimal places based on this? If this behavior were pushed more centrally, interpreting the column-specific format is cleaner than guessing a number and applying it everywhere.

Yeah thats a good call. Same behavior but tightens that up between client and server metadata.

Viral_Load_Assay/src/org/labkey/viral_load_assay/Viral_Load_Manager.java
```diff
@@ -75,6 +83,32 @@ public JSONObject getDefaultAssayMetadata(JSONObject meta){
         plate.put("hidden", true);
         resultsMeta.put("plate", plate);
 
+        // This overrides extjs default decimalPrecision of 2, only where specified by metadata from the server. Makes input
+        // forms respect formatString on column metadata. Decimals will still round to two places if formatString not set.
+        // Consider: Move somewhere more central?
+        TableInfo resultTable = QueryService.get().getUserSchema(ctx.getUser(), ctx.getContainer(), SchemaKey.fromString("assay.Viral_Loads.Viral_Load")).getTable("Data");
```
labkey-jeckels, line 89: This effectively hard-codes the assay design name of "Viral_Load". Should be fine as long as that's the only one they care about matching against.

labkey-jeckels (approved):

labkey-martyp:
Closing this PR to merge into 21.11 instead.  https://github.com/LabKey/ehrModules/pull/385

PR #385, opened 2022-05-12
Viral assay load decimal precision

labkey-martyp:
#### Rationale
ExtJS defaults number inputs to two decimal places. Currently this is not configurable for the viral load assays. This PR sets the decimal precision to match the formatString (if defined) for extjs decimal columns in viral load assay results. Default is still two. Setting formatString on column metadata will set the number of decimal places.

#### Related Pull Requests
https://github.com/LabKey/ehrModules/pull/383

#### Changes
* Default metadata for the assay results adds the decimalPrecision to any decimal or double values with a formatString defined

### DIFF
--- a/Viral_Load_Assay/src/org/labkey/viral_load_assay/Viral_Load_Manager.java
+++ b/Viral_Load_Assay/src/org/labkey/viral_load_assay/Viral_Load_Manager.java
@@ -16,6 +16,14 @@
 package org.labkey.viral_load_assay;
 
 import org.json.JSONObject;
+import org.labkey.api.data.ColumnInfo;
+import org.labkey.api.data.JdbcType;
+import org.labkey.api.data.TableInfo;
+import org.labkey.api.query.QueryService;
+import org.labkey.api.query.SchemaKey;
+import org.labkey.api.view.ViewContext;
+
+import java.text.DecimalFormat;
 
 /**
  * Created with IntelliJ IDEA.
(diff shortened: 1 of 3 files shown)


Viral_Load_Assay/src/org/labkey/viral_load_assay/Viral_Load_Manager.java
```diff
@@ -75,6 +83,32 @@ public JSONObject getDefaultAssayMetadata(JSONObject meta){
         plate.put("hidden", true);
         resultsMeta.put("plate", plate);
 
+        // This overrides extjs default decimalPrecision of 2, only where specified by metadata from the server. Makes input
+        // forms respect formatString on column metadata. Decimals will still round to two places if formatString not set.
+        // Consider: Move somewhere more central?
+        TableInfo resultTable = QueryService.get().getUserSchema(ctx.getUser(), ctx.getContainer(), SchemaKey.fromString("assay.Viral_Loads.Viral_Load")).getTable("Data");
```
labkey-jeckels: Carrying over my comment from the original PR, this is hard-coded with assumption that the assay design is named  Viral_Load. If that's true for all configurations that we care about, no worries.

labkey-jeckels (approved):

Viral_Load_Assay/src/org/labkey/viral_load_assay/Viral_Load_Manager.java
labkey-martyp: This is the only assay specific to this module, and the name is also hardcoded in xml config files.

Added the schema null check as discussed.

PR #398, opened 2022-07-20
Demographic cache logging on ETL cleanup

labkey-martyp:
#### Rationale
Message filling logs with lists of ids for animal records not cached during ETL, but first thing we do during async import is clear the cache.  This PR stops logging this message during ETL import.

#### Changes
* Don't log missing animal records from cache during ETL


labkey-jeckels (approved):
I assume the clearing that you're talking about is removing the IDs that are involved around line 255?

labkey-martyp:
> I assume the clearing that you're talking about is removing the IDs that are involved around line 255?

Yes that is correct

PR #412, opened 2022-09-01
Filter out dateDisabled Snomed codes

kollil:
Filter the datedisabled snomed codes out from the combo lists.


kollil:
@labkey-jeckels , This change is in the ehr modules, part of the core functionality. We have a situation at work where a medication needs to be disabled and should not be displayed in the combo lists. I just added the where clause to display only the ones that are not disabled (which is a reasonable approach to display the items). Let me know if this change can be applied to the core functionality or should I create an ONPRC version of this.

Thanks
Lakshmi

labkey-jeckels (changes requested):
The change to snomed_combo_list.sql looks fine. The PR also includes a new file, publicehr.template.xml, which seems unrelated and likely accidental. If you remove it, I'd be happy to approve.

labkey-jeckels (approved):

kollil:
@labkey-jeckels, 
I would like to know if this is the right place to make the change to filter out the date disabled snomeds. 
I see another js file, C:\onprc22.3\server\modules\ehrModules\ehr\resources\web\ehr\form\field\SnomedCombo.js , that's the xtype with query, filters and all other details. Should I be adding the filter here instead? I want to do this correctly because, this is the first time we have disabled a snomed since the column was  created few years ago but has never been used. 
I would like to exclude it in the drop down lists for the users to select. It has to be displayed in the history but not an option on the data entry screens to select.

Thanks
Lakshmi

labkey-jeckels:
@kollil I think this is the right change. I did some quick testing. When I set a disabled date, it disappeared from the Medications/Diet form's Treatment field drop down, but previously saved records that referenced it still found it.

kollil:
Thank you Josh. 
Hoping the changes are successful after the fix, but I still see the message saying there is an outstanding error. Is it it something that I need to be fixing before I hit the Squash & Merge button? Or, am I good to go with the build for the prod install.

labkey-jeckels:
@kollil I checked and the failures are unrelated to your change. I went ahead and merged on your behalf.

kollil:
Thank you @labkey-jeckels  :-)

PR #428, opened 2022-10-18
Allowing same day treatments

dnicolalde:
#### Rationale
Allow same day treatments on water monitoring system. 

#### Related Pull Requests
None

#### Changes
Adding a >= to the trigger.js script that compares start and end dates.


labkey-martyp (approved):
I don't see any issues with this.

PR #447, opened 2022-11-15
admin.jsp cleanup

labkey-adam:
#### Rationale
Several minor problems with this JSP

#### Changes
* Shouldn't compare unencoded vs. encoded values
* Shouldn't pass encoded content to tags (checkbox, in this case); that causes double encoding
* Unnecessary reference to `http://www.springframework.org/tags/form` prevented runtime recompilation
* Optimize selected lookups


labkey-martyp (approved):
I must have included that spring forms library when I was figuring out the form for this page and forgot to delete.

I didn't realize those attributes would be encoded automatically. Thanks for the cleanup.

labkey-adam:
Yeah, this is the exception to the rule. When outputting Strings in a JSP via `<%=myString%>` you always encode with `h()`... except when they're tag parameters. I kind of hate tags.

PR #454, opened 2022-11-23
Add DataEntry cache access to EHRService

labkey-martyp:
#### Rationale
In the related PR, the DataEntryManager cache needs to be accessed outside the EHR module in a trigger script helper.  This PR adds methods to EHRService to get and put from the data entry cache.

#### Related Pull Requests
* https://github.com/LabKey/onprcEHRModules/pull/550

#### Changes
* Add DataEntryManager cache get and put methods to EHRService


labkey-jeckels:
I'm not against exposing the cache, but I also don't see why it's necessary here. The code in ONPRC_EHRTriggerHelper.java is using its own class name in the key, so it won't be sharing the cache entries with the identical code in EHRTriggerHelper. I'd say that we should either consolidate the code and add methods like `getBloodDrawServicesMap()` to EHRService or use a separate cache in ONPRC_EHRTriggerHelper.java

PR #456, opened 2022-12-01
Adding room and cage as separate columns for Ext3 and Ext4

dnicolalde:
#### Rationale
We had a request from our user of the possibility to separate the location column into two columns to help sorting animal enclosures. There was already methods to get the room and cage, I modify the ext configuration to have it populate when the animalId get populated

#### Related Pull Requests
* 

#### Changes
Copy logic from location to have room and cage automatically populate when animalId is added

### DIFF
--- a/ehr/api-src/org/labkey/api/ehr/dataentry/SimpleFormSection.java
+++ b/ehr/api-src/org/labkey/api/ehr/dataentry/SimpleFormSection.java
@@ -41,6 +41,7 @@ public class SimpleFormSection extends AbstractFormSection
     private String _queryName;
 
     protected boolean _showLocation = false;
+    protected boolean _showRoomCage = false;
     /** Whether the grid has a button to edit the row in a popup dialog with a text label/form element layout for each field */
     protected boolean _allowRowEditing = true;
 
@@ -103,6 +104,10 @@ public class SimpleFormSection extends AbstractFormSection
         if (_showLocation)
         {
             keys.add(0, FieldKey.fromString("Id/curLocation/location"));
+        }
+        if (_showRoomCage){
+            keys.add(1, FieldKey.fromString("Id/curLocation/room"));
+            keys.add(2, FieldKey.fromString("Id/curLocation/cage"));
         }
 
         return keys;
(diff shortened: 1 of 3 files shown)


bbimber:
@dnicolalde If the use-case is sorting an Ext table, could you implement a custom sorting function that might accomplish the same goal? This could intelligently handle your room/cage formatting to ensure the expected order is returned?

dnicolalde:
At the moment the only column that gets populated is the location, which has a combination of room-cage. We are interested in separating the two to that the user can sort the Ext forms as they see adequate.

bbimber:
I understand what the PR is doing, but I'm exploring whether there's a different way to think about it. 

Even if you have two columns, my experience is that the client-side UI to on more than one column at a time can be cumbersome and it's easy to get the sort priority different from what you actually want. 

I can see how doing a naive lexicographic sort on the room/cage string would not give the user what they want. This said, it would be pretty simple (and maybe a good idea), to implement a JS sort function that returns the more human-expected result. As a default, I would guess you want the room portion with a natural-sort, followed by natural sort on the cage part of the string. I think the natural-sort is important (and possible in JS), so you get cage A1, A2...A10, rather than A1, A10, A2, etc. This is supposed to be how it sorts server-side, but this will not by default happen on the client. This would mean if the user tried to sort the existing column, it's much more likely they'd get a result that makes sense.

I'll add, the reference tables have a built-in idea of sort order for rooms, which we could leverage here if the natural order is something other than simple alphabetic/natural-sort. See ehr.rooms. 

Similarly, there's the concept of cage_position for cages in ehr.cages. The idea here was that the physical order of cages within a room doesnt always match an alphabetic order. If that is set up right, the database (and client sorting) could order them in whatever makes sense based on their actual locations.

@dnicolalde: What scenarios exist when the user wants something other than sorting by room/cage in order (assuming we got the "right" order)? I'm just curious what sorts would be performed besides room and cage in order?

labkey-jeckels:
@dnicolalde @keith-labkey I watched the recording of last week's call. I understand the new functionality now and see how it's opt-in behavior if any form wants it.

However, from what Jenny and Daniel said at just past the 48 minute mark, her desired behavior is already happening with the existing Location column. Daniel, would you be enabling this anywhere? If so, we can take it. But if not, then I'd rather not have unused code introduced. We could always go back and reopen the PR if we find a new use case in the future.

dnicolalde:
Hello, I did talked to the users and she mentioned that she will be okay with having the location as a single column. So we can close this PR.

labkey-jeckels:
OK, I'll go ahead and close it, but we can revisit if there are new requirements in the future.

PR #457, opened 2022-12-04
Animal Id Default Case Handling

labkey-martyp:
#### Rationale
By default, animal Ids are being forced to lower case in several UI components like Animal History, Animal Search and bulk entry in data entry forms. All but one center have overridden this behavior. This PR changes the default to not force any formatting changes, but provides a formatting function which can be overridden in PRC modules.

#### Related Pull Requests
* https://github.com/LabKey/wnprc-modules/pull/233

#### Changes
* Add overrides file with updated splitids function that does not force lower case and has tie in for custom formatting
* Make case insensitive animal history search the default case

### DIFF
--- a/ehr/resources/web/ehr/overrides.js
+++ b/ehr/resources/web/ehr/overrides.js
@@ -16,7 +16,7 @@ LDK.Utils.splitIds = function(subjectArray, unsorted, preserveDuplicates) {
         subjectArray = [];
     }
 
-    EHR.Utils.formatAnimalIds(subjectArray);
+    subjectArray = EHR.Utils.formatAnimalIds(subjectArray);
 
     if (subjectArray.length > 0) {
 
--- a/ehr/resources/web/ehr/overrides.js
+++ b/ehr/resources/web/ehr/overrides.js
@@ -1,0 +1,34 @@
+
+
+LDK.Utils.splitIds = function(subjectArray, unsorted, preserveDuplicates, preserveCase) {
+    if (!subjectArray){
+        return [];
+    }
+
+    subjectArray = Ext4.String.trim(subjectArray);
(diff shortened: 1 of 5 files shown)


ehr/resources/web/ehr/utils.js
```diff
@@ -626,6 +626,13 @@ EHR.Utils = new function(){
                 minWidth: 60,
                 border: includeBorder
             };
+        },
+
+        /** Override this function to apply custom formatting to ids entered into UI components like animal history, animal search,
```
labkey-jeckels: "Custom formatting" makes me think of bold, italics, or other HTML-based formatting. The intent is to deal with casing, hyphens, etc, right? Updating the comment to make that clear would be good.

labkey-jeckels (approved):

labkey-jeckels (approved):

labkey-jeckels (approved):

ehr/resources/web/ehr/overrides.js
```diff
@@ -0,0 +1,32 @@
...
+    if (subjectArray){
+        subjectArray = subjectArray.split(';');
+    }
+    else {
+        subjectArray = [];
+    }
+
+    subjectArray = EHR.Utils.formatAnimalIds(subjectArray);
+
+    if (subjectArray.length > 0) {
```
bbimber: this is picky, but shouldnt this test greater than 1 rather than zero? do we actually need this check at all?

labkey-martyp: Yeah true I updated to greater than 1. The check just saves a little processing.

ehr/resources/web/ehr/utils.js
labkey-martyp: comment updated

PR #458, opened 2022-12-05
Groups and alias queries and reports

labkey-martyp:
#### Rationale
These are some common queries, metadata and views around groups and aliases. These will be part of the base EHR but are overridable in custom modules.

#### Changes
* Add alias dataset and related metadata
* Add animal group views used in group details
* Add ensure single flag trigger and setup only when using ehr_lookups.flag_values


labkey-jeckels (approved):
Are there any other copies of these resources in other PRC-specific modules that can be deleted because they're identical to these?

PR #463, opened 2022-12-12
More Base EHR Updates

labkey-martyp:
#### Rationale
Copy housing form type and sources demographic provider to EHR. Close Protocol Assignments on death/departure for centers not using projects.

#### Related Pull Requests
* https://github.com/LabKey/johnsHopkinsEHRModules/pull/17

#### Changes
* Copy SourceDemographicsProvider to EHR
* Add Protocol Assignments to list of dataset records to close


ehr/api-src/org/labkey/api/ehr/dataentry/forms/HousingFormType.java
```diff
@@ -0,0 +1,72 @@
...
+import org.labkey.api.view.template.ClientDependency;
+
+import java.util.Arrays;
+
+/**
+ * User: bimber
+ * Date: 7/29/13
+ * Time: 5:03 PM
+ */
+public class HousingFormType extends UnsaveableTask
```
labkey-jeckels: Why are you adding this to core module if nobody is using it? Could we make it easier to leverage by letting each center plug in its down AnimalDetails section instead of having to create a new class?

ehr/api-src/org/labkey/api/ehr/demographics/SourceDemographicsProvider.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * User: bimber
+ * Date: 7/9/13
+ * Time: 10:12 PM
+ */
+public class SourceDemographicsProvider extends AbstractListDemographicsProvider
```
labkey-jeckels, line 30: Similar question - why copy this to the EHR module if nobody references it? I understand we want to more the core EHR module more functional, but I'd rather prioritize the items where we have a concrete need instead of pulling misc classes in the hopes they might be useful for a hypothetical future customer. Or did I miss usage in JHU?

labkey-jeckels (changes requested):
I'm not sure

ehr/api-src/org/labkey/api/ehr/dataentry/forms/HousingFormType.java
labkey-martyp: I removed it. I was originally going to use it but ended up not. Thought it would be ok to leave for base EHR but it is not being used so I'll remove it for now and think about a refactor.

ehr/api-src/org/labkey/api/ehr/demographics/SourceDemographicsProvider.java
labkey-martyp: This is being used in JHU as a registered demographics provider.

labkey-jeckels (approved):

PR #465, opened 2022-12-14
Test fix.

labkey-sweta:
#### Rationale
Test fix for 
TNPRC_EHRTest.testAnimalHistoryReports
There appears to be an error in the report: SNP

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/ehr/test/src/org/labkey/test/pages/ehr/ParticipantViewPage.java
+++ b/ehr/test/src/org/labkey/test/pages/ehr/ParticipantViewPage.java
@@ -386,7 +386,7 @@ public class ParticipantViewPage<EC extends ParticipantViewPage.ElementCache> ex
                 scrollIntoView(_el);
                 doAndWaitForRepeatedPageSignal(_el::click, REPORT_TAB_SIGNAL, Duration.ofSeconds(2));
                 _ext4Helper.waitForMaskToDisappear(30000);
-                activeReportPanelContainer.waitForElement(getDriver(), 10000);
+                activeReportPanelContainer.waitForElement(getDriver(), 20000);
                 sleep(5000);
 
             }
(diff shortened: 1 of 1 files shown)


labkey-tchad (changes requested):
Looks like `doAndWaitForRepeatedPageSignal` isn't respecting the timeout (the error says that it failed after waiting for 10 seconds). This means that this increase wouldn't have helped and that there's a problem with the helper.

ehr/test/src/org/labkey/test/pages/ehr/ParticipantViewPage.java
```diff
@@ -384,11 +384,10 @@ public void select()
             {
                 log("Selecting Report: " + getLabel());
                 scrollIntoView(_el);
-                doAndWaitForRepeatedPageSignal(_el::click, REPORT_TAB_SIGNAL, Duration.ofSeconds(2));
+                doAndWaitForRepeatedPageSignal(_el::click, REPORT_TAB_SIGNAL, Duration.ofSeconds(3));
```
labkey-tchad: Actually that `Duration` parameter isn't the timeout, that's how long it waits for the page signal to stop updating. This needs to use the overload of this method that takes a custom timeout.
```suggestion
                doAndWaitForRepeatedPageSignal(_el::click, REPORT_TAB_SIGNAL, Duration.ofSeconds(2), longWait());
```

labkey-tchad (changes requested):

ehr/test/src/org/labkey/test/pages/ehr/ParticipantViewPage.java
labkey-sweta: Thank you

labkey-tchad (approved):
Seems to have done the trick

PR #471, opened 2022-12-17
Close animal group records on death/departure

labkey-martyp:
#### Rationale
Add animal group members to list of records to close on death or departure.

#### Related Pull Requests
* https://github.com/LabKey/johnsHopkinsEHRModules/pull/22

#### Changes
* ScriptHelper datasetsToClose updated

### DIFF
--- a/ehr/resources/scripts/ehr/ScriptHelper.js
+++ b/ehr/resources/scripts/ehr/ScriptHelper.js
@@ -85,7 +85,7 @@ EHR.Server.ScriptHelper = function(extraContext, event, EHR){
      */
 
     var scriptOptions = {
-        datasetsToClose: ['Assignment', 'Cases', 'Housing', 'Treatment Orders', 'Notes', 'Problem List', 'Protocol Assignments'],
+        datasetsToClose: ['Assignment', 'Cases', 'Housing', 'Treatment Orders', 'Notes', 'Problem List', 'Protocol Assignments', 'Animal Group Members'],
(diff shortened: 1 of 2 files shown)


ehr/resources/scripts/ehr/ScriptHelper.js
```diff
@@ -85,7 +85,7 @@ EHR.Server.ScriptHelper = function(extraContext, event, EHR){
      */
 
     var scriptOptions = {
-        datasetsToClose: ['Assignment', 'Cases', 'Housing', 'Treatment Orders', 'Notes', 'Problem List', 'Protocol Assignments'],
+        datasetsToClose: ['Assignment', 'Cases', 'Housing', 'Treatment Orders', 'Notes', 'Problem List', 'Protocol Assignments', 'Animal Group Members'],
```
labkey-jeckels, line 88: Do all centers have all of these tables? I see that we don't blow up if one is absent, but I don't think we really want to be logging about it either.

https://github.com/LabKey/ehrModules/blob/e27cc4a088c961488f5e3750ef1b1cb0fe8f53de/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java#L199

labkey-martyp: Yeah it just continues if the dataset is not found. I think this is one that should be part of the base EHR. I removed the log message.

labkey-jeckels (approved):

PR #473, opened 2022-12-19
Reverting change to allow one day treatment

dnicolalde:
#### Rationale
Back in Oct, 2022 we made a change to the code. This is causing some problems and it was a mistake

#### Related Pull Requests
* #428

#### Changes
* Removing = from the >= statement


labkey-keith:
@dnicolalde When you have a PR for the EHR repository please include Marty and Josh as reviewers. Thanks!

labkey-jeckels (approved):

PR #476, opened 2022-12-21
Ticket 46941: Data entry grid view height recalculate

labkey-martyp:
#### Rationale
Grid views in data entry do not always correctly re-set height as the contents change or are updated. This update just ensures the grid view height is set to the height of it's body to ensure all contents are displayed. Covers initial load, adding/removing rows and multi-line text cell edits

#### Changes
* Add resize height function
* Add updateLayout override to adjust height of grid view to contents
* Add resize after validation complete

### DIFF
--- a/ehr/resources/web/ehr/grid/Panel.js
+++ b/ehr/resources/web/ehr/grid/Panel.js
@@ -93,7 +93,7 @@ Ext4.define('EHR.grid.Panel', {
     heightResize: false, // avoid infinite loop
     updateLayout: function(options){
         const view = this.getView();
-        if (view.rendered && view.body && !this.heightResize) {
+        if (view.rendered && view.body && view.body.dom && !this.heightResize) {
             this.heightResize = true;
             view.setHeight(view.body.getHeight());
         }
--- a/ehr/resources/web/ehr/grid/Panel.js
+++ b/ehr/resources/web/ehr/grid/Panel.js
@@ -60,13 +60,6 @@ Ext4.define('EHR.grid.Panel', {
 
         this.getSelectionModel().on('selectionchange', this.handleSectionChangeEvent, this);
 
-        // Ensure gridview height updated on long text cell edits
-        if (this.plugins.length > 0) {
-            this.plugins[0].on('edit', this.resizeHeight, this);
-            this.plugins[0].on('beforeedit', this.resizeHeight, this);
-            this.plugins[0].on('canceledit', this.resizeHeight, this);
-        }
-
         // the intention of the following is to avoid redrawing the entire grid, which is expensive, when we have
         // single row changes, or more importantly single row changes that only involve validation/tooltip error message differences
         this.on('storevalidationcomplete', this.onStoreValidationComplete, this, {buffer: 100, delay: 20});
@@ -95,6 +88,20 @@ Ext4.define('EHR.grid.Panel', {
     },
 
     pendingChanges: {},
+
+    getEditingPlugin: function(){
+        const plugin = Ext4.create('LDK.grid.plugin.CellEditing', {
+            pluginId: this.editingPluginId,
+            clicksToEdit: this.clicksToEdit
+        });
+
+        // Ensure gridview height updated on long text cell edits
+        plugin.on('edit', this.resizeHeight, this);
+        plugin.on('beforeedit', this.resizeHeight, this);
+        plugin.on('canceledit', this.resizeHeight, this);
+
+        return plugin;
+    },
 
     resizeHeight: function(){
         // A bit of a hack but there are some cases where the gridview does not resize height to match its contents. Called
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):

bbimber:
Hey - 

I believe this bug exists, but have two questions: 

1) do you have a sense on what kind of overhead all this re-rendering causes? Does resizing the height trigger a full redo of the layout, meaning this might introduce a lot more repeat doLayout calls? If true, that's potentially a non-trivial change. If you're trying the grid with a handful of rows, this probably isnt even noticeable, and would work fine. However, when working on grids/forms with a lot of rows (perhaps daily observations / rounds, and/or some other workflow where entry involves whole rooms), then this might matter a lot more. I dont know what bookkeeping is associated with changing the view height, for example. If there is non-trivial work, is there a low-cost way to check the current height and compare to the new value to limit when you call this?

2) In general I appreciate trying to limit the scope of the change (i.e. EHR vs. LDK). Do you have a sense on whether the problem of rows getting out-of-sync with height is EHR-specific (perhaps due to the background validation), or is it a more general issue Ext4 grids can have? My sense is that it was caused by text-wrapping, and perhaps due to Ext4's layout/height calc not handling this optimally. But it was a long time ago when I really looked at this.

labkey-martyp:
> Hey -
> 
> I believe this bug exists, but have two questions:
> 
> 1. do you have a sense on what kind of overhead all this re-rendering causes? Does resizing the height trigger a full redo of the layout, meaning this might introduce a lot more repeat doLayout calls? If true, that's potentially a non-trivial change. If you're trying the grid with a handful of rows, this probably isnt even noticeable, and would work fine. However, when working on grids/forms with a lot of rows (perhaps daily observations / rounds, and/or some other workflow where entry involves whole rooms), then this might matter a lot more. I dont know what bookkeeping is associated with changing the view height, for example. If there is non-trivial work, is there a low-cost way to check the current height and compare to the new value to limit when you call this?
> 2. In general I appreciate trying to limit the scope of the change (i.e. EHR vs. LDK). Do you have a sense on whether the problem of rows getting out-of-sync with height is EHR-specific (perhaps due to the background validation), or is it a more general issue Ext4 grids can have? My sense is that it was caused by text-wrapping, and perhaps due to Ext4's layout/height calc not handling this optimally. But it was a long time ago when I really looked at this.

1. Yeah initially I tried only doing this when view and view body heights didn't match but there are some cases like bulk update that are editing the grid directly where the heights match on the components but not in the display. Calling setHeight does the bookkeeping to trigger the layout update. That being said, as far as I can tell this doesn't increase the number of actual layout updates on the component. The setHeight does only that then refers to the parent to call updateLayout which loops back through this component to call the normal updateLayout. At least as far as AbstractComponent.updateLayout calls they are the same. But it's a good point to try this out on a larger set of records. I'll do that before merging.
2. This is definitely due to some manipulation of data after the grid render and possibly during validation. I didn't dig too deep into trying to fix an individual form, instead doing the more generic approach to make the grid view more reactive to the contents. I'm not sure it's EHR specific other than there are a lot of varying requirements and customizations.

bbimber:
OK, thanks @labkey-martyp. After xmas I might take a look into this for non-EHR grids. The ones I know about a a lot less demanding than EHR, but they might encounter this too. If that turns out to be the case, I might suggest pushing this into the base LDK class.

ehr/resources/web/ehr/grid/Panel.js
```diff
@@ -89,18 +89,18 @@ Ext4.define('EHR.grid.Panel', {
...
+    heightResize: false, // avoid infinite loop
+    updateLayout: function(options){
+        const view = this.getView();
+        if (view.rendered && view.body && !this.heightResize) {
+            this.heightResize = true;
+            view.setHeight(view.body.getHeight());
+        }
+        else {
+            this.heightResize = false;
+            this.callParent();
```
bbimber: If this method has arguments, should this be: this.callParent(arguments)?

bbimber (changes requested):
That's great you were able to simplify this into a relatively discrete change in the grid component alone. Minor comment above.

ehr/resources/web/ehr/grid/Panel.js
bbimber: actually - so this only reaches callParent() when heightResize=false? So if heightResize=true, then we never call the parent class? Is that a good thing?

labkey-martyp: Good call on the arguments. Updated that.

Ultimately setHeight calls setSize which calls updateLayout on the component again, which is ultimately why we need the heightResize flag to avoid an infinite loop.  https://docs.sencha.com/extjs/4.2.1/source/AbstractComponent.html#Ext-AbstractComponent-method-setSize

labkey-martyp:
@bbimber Let me know if you have any more thoughts on this Ben. Otherwise it needs your approval to get merged in.

bbimber:
@labkey-martyp I wont have a any to do any real investigation until next week. If you want to merge, i just approved. I am going to do a little looking to see if i can repro this and if it makes sense in LDK, but I dont have any specific comments beyond what I already added. I dont understand the infinite loop issue enough without digging into code myself.

bbimber (approved):

PR #493, opened 2023-01-04
Issue 47002: EHR Kinship Calculations are not consistent and incorrect across servers

labkey-martyp:
#### Rationale
Not sure if this is due to a change in the kinship2 library, but the kinship2 kinship function requires pedigree in a specific order to properly calculate all kinships across a pedigree. This PR does some refactoring in populateKinship.r to use more kinship2 built-in functionality. Additionally, an optional step added to kinship calculations to do some broad validation of coefficients across demographic family queries.

- This update will require EHR users to upgrade the related R packages and add an additional R package, dplyr

#### Related Pull Requests
* https://github.com/LabKey/onprcEHRModules/pull/575

#### Changes
* Update populateKinship.r to use kinship2 fixParents and pedigree functions to prepare data for kinship calc.
* Update populateInbreeding to only calculate inbreeding for passed in Ids from pedigree.sql
* Add optional setting to do extra kinship validation with genetic calculations
* Add extra kinship validation to verify against demographicsFamily, demographicsSiblings and demographicsOffspring
* Add demographicsSibling and demographicsOffspring queries to EHR

### DIFF
--- a/ehr/resources/pipeline/kinship/populateInbreeding.r
+++ b/ehr/resources/pipeline/kinship/populateInbreeding.r
@@ -31,8 +31,7 @@ is.na(allPed$Gender)<-which(allPed$Gender=="")
 df <- data.frame(id=as.character(allPed$Id), 'id parent1'=allPed$Dam, 'id parent2'=allPed$Sire, stringsAsFactors=FALSE);
 colnames(df)<-c("id", "id parent1", "id parent2")
 
-originalIds <-as.data.frame(df[,1])
-colnames(originalIds) <- c("id")
+originalIds <-as.data.frame(df[,1,drop=FALSE])
 
 #this is a function in the pedigree package designed to add missing parents to the dataframe
 #see pedigree package documentation for more detail
--- a/ehr/resources/pipeline/kinship/populateInbreeding.r
+++ b/ehr/resources/pipeline/kinship/populateInbreeding.r
@@ -13,6 +13,7 @@ options(error = dump.frames);
 library(pedigree);
 library(getopt);
 library(Matrix);
+library(dplyr)
 
 spec <- matrix(c(
 'inputFile', '-f', 1, "character"
@@ -30,6 +31,9 @@ is.na(allPed$Gender)<-which(allPed$Gender=="")
 df <- data.frame(id=as.character(allPed$Id), 'id parent1'=allPed$Dam, 'id parent2'=allPed$Sire, stringsAsFactors=FALSE);
 colnames(df)<-c("id", "id parent1", "id parent2")
 
+originalIds <-as.data.frame(df[,1])
+colnames(originalIds) <- c("id")
+
 #this is a function in the pedigree package designed to add missing parents to the dataframe
 #see pedigree package documentation for more detail
 df <- add.Inds(df);
@@ -40,6 +44,9 @@ df <- df[order(ord),]
 ib = calcInbreeding(df);
 
 newRecords <- data.frame(Id=as.character(df$id), coefficient=ib, stringsAsFactors=FALSE);
+
+#only calculate inbreeding for Ids at the center
+newRecords <- dplyr::filter(newRecords, Id %in% originalIds$id)
 
 # write TSV to disk
 print("Output table:");
--- a/ehr/resources/pipeline/kinship/populateKinship.r
+++ b/ehr/resources/pipeline/kinship/populateKinship.r
@@ -15,6 +15,7 @@ library(methods);
 library(kinship2);
 library(getopt);
 library(Matrix);
+library(dplyr);
 
 spec <- matrix(c(
 #'containerPath', '-c', 1, "character",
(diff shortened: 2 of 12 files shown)


ehr/resources/queries/ehr/kinshipSummary.sql
```diff
@@ -3,19 +3,6 @@
...
-SELECT
-  k.Id2 as Id,
-  k.Id as Id2,
-  k.coefficient
-
-FROM ehr.kinship k
-WHERE k.Id != k.Id2
\ No newline at end of file
+SELECT * FROM ehr.kinship
+WHERE Id != Id2
```
bbimber, line 8: isnt this defeating the entire point of this script???

ehr/resources/queries/study/demographicsOffspring.sql
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) 2015-2016 LabKey Corporation
+ *
+ * Licensed under the Apache License, Version 2.0: http://www.apache.org/licenses/LICENSE-2.0
+ */
+SELECT
+    d.id as Id,
+    d2.id  AS Offspring,
+    d2.birth AS date,
+    d2.sire,
```
bbimber, line 10: i dont think these fields are accurate at ONPRC? what's the rationale for adding this script to EHR

ehr/resources/queries/study/demographicsSiblings.sql
```diff
@@ -0,0 +1,29 @@
...
+        END AS Relationship,
+
+    d2.id  AS Sibling,
+
+    d2.id.parents.dam AS SiblingDam,
+    d2.id.parents.sire AS SiblingSire,
+    d1.qcstate
+
+FROM study.Demographics d1
+JOIN study.Demographics d2 ON ((d2.id.parents.sire = d1.id.parents.sire OR d2.id.parents.dam = d1.id.parents.dam) AND d1.id != d2.id)
```
bbimber: same comments about demographics.dam and .sire

ehr/resources/pipeline/kinship/populateInbreeding.r
```diff
@@ -30,6 +31,9 @@ is.na(allPed$Gender)<-which(allPed$Gender=="")
 df <- data.frame(id=as.character(allPed$Id), 'id parent1'=allPed$Dam, 'id parent2'=allPed$Sire, stringsAsFactors=FALSE);
 colnames(df)<-c("id", "id parent1", "id parent2")
 
+originalIds <-as.data.frame(df[,1])
```
bbimber: originalIds <-df[,1,drop=FALSE]

does does this. drop=FALSE preserves the single-column vector as a data.frame. you dont need to set colnames() then either.

ehr/resources/pipeline/kinship/populateInbreeding.r
```diff
@@ -41,6 +45,9 @@ ib = calcInbreeding(df);
 
 newRecords <- data.frame(Id=as.character(df$id), coefficient=ib, stringsAsFactors=FALSE);
 
+#only calculate inbreeding for Ids at the center
+newRecords <- dplyr::filter(newRecords, Id %in% originalIds$id)
```
bbimber, line 48: this doesnt really matter that much but you could do this without dplyr.

ehr/resources/pipeline/kinship/populateKinship.r
```diff
@@ -35,69 +36,35 @@ allPed$Species <- as.character(allPed$Species)
...
-        temp.tri=data.frame(Id=colnames(temp.kin)[sparse.kin@i+1],Id2=colnames(temp.kin)[sparse.kin@j+1],coefficient=sparse.kin@x,stringsAsFactors=FALSE)
-        newRecords=rbind(newRecords,temp.tri)
-    }
+    allRecordsForSpecies <- allPed[allPed$Species == species,]
+
+    # Add missing parents for accurate kinship calculations
+    fixedRecords <- with(allRecordsForSpecies, fixParents(id = Id, dadid = Sire, momid = Dam, sex = Gender))
+
+    # Kinship is expecting records to be sorted IAW it's own pedigree function
+    recordsForSpecies <- with(fixedRecords, pedigree(id=id,dadid=dadid,momid=momid,sex=sex,missid=0))
```
bbimber, line 50: The original rationale in splitting the data by kinship's makefamid was to reduce the data volume and calculation time. how big a dataset have you used here? this was written in probably 2010, so it's possible either modern servers and/or improvements to kinship() mean that's no longer necessary. It's certainly probably better if we can compute everything in one call per species, so that we dont get false-negatives from relatives being split into different families by makefamid. do oyu have the logs from a previously successful ONPRC kinship run, to see how many inferred families there were, and how big those families were?

bbimber (changes requested):
see comments. overall i think there's some important and useful changes here but i have questions on some specifics.

ehr/resources/pipeline/kinship/populateKinship.r
labkey-martyp: This actually considerably speeds up the calculations, at least on my laptop and did not run into issues with resources. I suspect any servers will have more resources than my laptop. I tried out a pretty complex colony dataset with about 40k animals and wide variation of family sizes.

The part that can have performance issues is the optional validation I added that just goes through the pedigree and verifies there are reasonable kinship coefficients for every first or second degree relation. This is likely going to be a one time run for each center to see if there are any issues.

ehr/resources/pipeline/kinship/populateInbreeding.r
labkey-martyp: Yeah I figured you could, dplyr is just a proven way to do it efficiently so I went with that solution. Also using dplyr in populateKinship.

ehr/resources/pipeline/kinship/populateInbreeding.r
labkey-martyp: Nice, works great. Thanks.

ehr/resources/queries/study/demographicsSiblings.sql
labkey-martyp: Same comment as above. ONPRC has this overridden, this is added for other PRCs.

ehr/resources/queries/study/demographicsOffspring.sql
labkey-martyp: ONPRC has this overridden so it won't affect them. This is added for other PRCs.

ehr/resources/queries/ehr/kinshipSummary.sql
labkey-martyp: I'm not sure how it's defeating the point of the script. It's not really necessary other than there are reports in some of the PRCs that are using this query so I don't want to pull the rug out from under those. Before this change it was causing duplicate rows with the new calculations.

I'll file tickets to get those usages refactored to just use ehr.kinship but that will likely not be in this PR.

bbimber: Your answer might be saying this: but the content of what's stored in the database changed, I take it?

The original kinship data stored in the table included each pair of IDs represented once. Did that change? The point of this union was to give one single query, viewable in a LK dataframe, where the user could filter on ID=XX, and get all the results. If the DB table is actually storing two rows for each pair (i.e. ID1=A, ID2=B, and then ID1=B and ID2=A), then you are right here.

I also agree that it would be good to refactor out this SQL script if that is actually the case. I think you'd need ot also update the rows in the ehr.reports table (or at least check them), in case they reference this query directly.

ehr/resources/queries/study/demographicsOffspring.sql
bbimber: hmm, ok. I wonder if ONPRC should use their customizer to add calculated fields to study.demographics for sire/dam. that's a separate question from here.

ehr/resources/pipeline/kinship/populateKinship.r
bbimber: OK, long as it's tested. Splitting species is a possible source for inaccurate/missing coefficients too, so its nice to remove it. Do you have any kind of sense how much memory it actually is using for those 40K? In the past this was often run on a pipeline server, not the webserver. But I believe ONPRC and maybe others are running this job on the webserver itself. Maybe that's not a real issue in practice, but it historically was a sizable task.

ehr/resources/queries/study/demographicsSiblings.sql
```diff
@@ -0,0 +1,29 @@
...
+ * Copyright (c) 2015-2018 LabKey Corporation
+ *
+ * Licensed under the Apache License, Version 2.0: http://www.apache.org/licenses/LICENSE-2.0
+ */
+SELECT
+
+    d1.id,
+
+    CASE
+        WHEN (COALESCE(d1.id.parents.sire, '') = COALESCE(d2.id.parents.sire, '') and COALESCE(d1.id.parents.dam, '') = COALESCE(d2.id.parents.dam, '') AND COALESCE(d1.id.parents.sire, '')!='' AND COALESCE(d1.id.parents.dam, '')!='')
```
bbimber: Doesnt LabKey SQL have a built-in function that does an equals test accounting for NULL? I thought I saw this before.

bbimber (changes requested):
@labkey-martyp: thanks for the changes. my main comment on functionality is to verify that either the data in the table represents both combinations (i.e. AnimalA/AnimalB and AnimalB/AnimalA), or update kinshipSummary.sql.

ehr/resources/pipeline/kinship/populateKinship.r
labkey-martyp: Well it pretty much pegged out my memory at about 40GB above steady state, I'm assuming it was writing out to disk at the point. But that was only for a few seconds. I don't know how to test it out on everyone's servers other than to just commit it and let everyone try it out.

ehr/resources/queries/study/demographicsOffspring.sql
labkey-martyp: Yeah ONPRC keeps the dam and sire in the birth dataset and adds a parents node to the animal record in their customizer, so querying dam and sire from demographics is more like d.parents.dam.

ehr/resources/queries/ehr/kinshipSummary.sql
labkey-martyp: Yes that's all true. Kinship coefficient calculates for Ids in both directions. That's just the default way kinship2 is calculating the pedigree then the kinship, so don't need to do the union reversing the rows.  And yes the animal history kinship report will need to be updated to just use ehr.kinship for any centers using ehr.kinshipSummary.

ehr/resources/queries/study/demographicsOffspring.sql
bbimber: This is tangential but it's important to be precise on the data. the birth dataset's dam and sire are what's entered at birth (observed data). later, genetic testing can happen (potentially more than one result/record), and those take precedence (hence why demographicsParentage.sql exists). 

One big theme on how EHR is designed is the fact that data changes, even things one might generally assume dont (like your parents).

For here I was musing that maybe there's value in adding calculated SQL fields for demographics.dam and demographics.sire, which might let more core code that assumes those fields exist work. but that's beyond the scope here.

ehr/resources/pipeline/kinship/populateKinship.r
bbimber: ok

ehr/resources/queries/ehr/kinshipSummary.sql
bbimber: OK.

ehr/resources/queries/study/demographicsSiblings.sql
labkey-martyp: There is not. It's a frustration. 

I'm not a real big fan of doing it this way, seems like more work doing the coalescing each time than just doing a null check once. I'll look into changing that.

bbimber: what about ISEQUAL()? https://www.labkey.org/Documentation/wiki-page.view?name=labkeysql

i dont have code in front of me, but i think I remember writing a lot of EHR SQL with complex equals/null checks. but if that LK function works, it seems like it'd be useful here.

labkey-martyp: Oh yeah, guess I was thinking of t-sql. That looks useful.

bbimber (approved):

ehr/resources/queries/study/demographicsSiblings.query.xml
```diff
@@ -0,0 +1,44 @@
+<query xmlns="http://labkey.org/data/xml/query">
+    <description>Siblings of Any Animal</description>
+    <metadata>
+        <tables xmlns="http://labkey.org/data/xml">
+            <table tableName="demographicsSiblings" tableDbType="NOT_IN_DB">
+                <description>Siblings of Any Animal</description>
+                <columns>
+                    <column columnName="Id">
+                        <isKeyField>true</isKeyField>
+                        <!--<isHidden>true</isHidden>-->
```
labkey-jeckels: Trim this commented code?

ehr/resources/web/ehr/panel/GeneticCalculationSettingsPanel.js
```diff
@@ -24,6 +24,10 @@ Ext4.define('EHR.panel.GeneticCalculationSettingsPanel', {
                 xtype: 'checkbox',
                 fieldLabel: 'Is Enabled?',
                 itemId: 'enabled'
+            },{
+                xtype: 'checkbox',
+                fieldLabel: 'Kinship validation?',
```
labkey-jeckels, line 29: Can we add a tooltip or other helper to give users a little more context about what this does?

ehr/src/org/labkey/ehr/pipeline/GeneticCalculationsImportTask.java
```diff
@@ -272,6 +283,187 @@ else if (kinshipTable.getSqlDialect().isPostgreSQL())
...
+                    else
+                    {
+                        throw new IllegalStateException("Kinship validation failed for Id: " + id + ", Id2: " + id2 + ". Relation: " + relation + ". Kinship coefficent not found");
+                    }
+                }
+            }
+        }
+        catch (SQLException e)
+        {
+            throw new RuntimeException(e);
```
labkey-jeckels: Better to make this a RuntimeSQLException

ehr/src/org/labkey/ehr/pipeline/GeneticCalculationsImportTask.java
```diff
@@ -272,6 +283,187 @@ else if (kinshipTable.getSqlDialect().isPostgreSQL())
...
+        if (demographicsTable == null)
+        {
+            throw new IllegalStateException("Could not find query 'Demographics' in study schema");
+        }
+
+        SimpleFilter filter = new SimpleFilter(FieldKey.fromParts("Id"), List.of(id, id2), CompareType.IN);
+        TableSelector demographicsTs = new TableSelector(demographicsTable, PageFlowUtil.set("Id", "species"), filter, null);
+
+        String species = null;
+        try(Results results = demographicsTs.getResults())
```
labkey-jeckels, line 386: If this ends up being too slow for groups that want to validate, we could validate every X rows from demographics, or Y% of total records. Checking 10% of the animals would likely turn up if there are systemic problems, and of course run a lot faster. No need to do this proactively.

labkey-jeckels (approved):

ehr/resources/queries/study/demographicsSiblings.query.xml
labkey-martyp: Removed

ehr/resources/web/ehr/panel/GeneticCalculationSettingsPanel.js
labkey-martyp: Added a tooltip

ehr/src/org/labkey/ehr/pipeline/GeneticCalculationsImportTask.java
labkey-martyp: Updated

ehr/src/org/labkey/ehr/pipeline/GeneticCalculationsImportTask.java
labkey-martyp: I ended up refactoring based on our convo to get much faster validation. Still may take a while depending on the size of the pedigrees and server resources, but much faster now.

PR #519, opened 2023-02-05
Additional template and script property functionality

labkey-martyp:
#### Rationale
Make it easier to update existing section and form templates by adding capability to overwrite an existing template through the UI (instead of just throwing unique constraint error). Add ability to register trigger script properties in the java module file.

#### Related Pull Requests
* https://github.com/LabKey/johnsHopkinsEHRModules/pull/44

#### Changes
* Add registration mechanism to register trigger script options in the module and getter in java trigger helper.
* Update ScriptHelper to apply registered script options.
* For both form and section templates, first check if the template name already exists for that form or section. If so, then confirm with user to overwrite it with the current form/section values. On yes, delete old template and insert new template. Other scenarios are the same behavior.

### DIFF
--- a/ehr/api-src/org/labkey/api/ehr/EHRService.java
+++ b/ehr/api-src/org/labkey/api/ehr/EHRService.java
@@ -211,6 +211,10 @@ abstract public class EHRService
 
     abstract public void registerDefaultFieldKeys(String schemaName, String queryName, List<FieldKey> keys);
 
+    abstract public void registerTriggerScriptOption(String name, Object value);
+
+    abstract public Map<String, Object> getTriggerScriptOptions();
+
     public enum FORM_SECTION_LOCATION
     {
         Header(),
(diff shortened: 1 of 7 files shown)


bbimber:
No real concerns with the functionality and it seems reasonable. However: would it make sense to have at least some level of automated testing on new features? EHR's testing could always stand to be improved. That EHRService registry thing might be amenable to a java integration or unit test?
'
Bigger picture on testing: I gather there was a move away from "generic" EHR tests, in favor of only having the center-specific tests. I get the reasoning since EHR requires so much context. However, if there is now a move to push more into core EHR again, maybe there needs to be a slightly different approach? Perhaps there should be a simple TestEHR module, which might contain basic data and exercise some of these generic functions?

labkey-martyp:
> No real concerns with the functionality and it seems reasonable. However: would it make sense to have at least some level of automated testing on new features? EHR's testing could always stand to be improved. That EHRService registry thing might be amenable to a java integration or unit test? ' Bigger picture on testing: I gather there was a move away from "generic" EHR tests, in favor of only having the center-specific tests. I get the reasoning since EHR requires so much context. However, if there is now a move to push more into core EHR again, maybe there needs to be a slightly different approach? Perhaps there should be a simple TestEHR module, which might contain basic data and exercise some of these generic functions?

We are saving some budget and time to do automated testing on these features. We should start on that in the next week or so.

It does sound like a good idea to have some core tests, right now everything is spread across the center's modules. We can look into that as part of the bigger effort in updating EHR. At least put core testing for the new UI/features.

labkey-jeckels (approved):

PR #529, opened 2023-02-10
Fix WNPRC_PurchasingTest.testEmailCustomizationLineItemUpdate

XingY:
#### Rationale
ehr_purchasing.purchasingRequests has a custom query definition in wnprc_purchusing that adds special calculated columns. The email notifications sent are dependent on those calculated columns, which aren't returned by updateRows using prepared statement. This PR flag all ehr_purchasing tables to use the old row-by-row updates, instead of prepared statement.

TC failure:
https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_EhrPostgres/2274993?buildTab=tests&status=failed

#### Related Pull Requests
* https://github.com/LabKey/ehrModules/pull/529
* https://github.com/LabKey/wnprc-modules/pull/251

#### Changes
* update updateRows parameters to match updated signature
* don't use data iterator to update ehr_purchasing tables

### DIFF
--- a/EHR_Purchasing/src/org/labkey/ehr_purchasing/EHR_PurchasingTable.java
+++ b/EHR_Purchasing/src/org/labkey/ehr_purchasing/EHR_PurchasingTable.java
@@ -1,38 +1,18 @@
 package org.labkey.ehr_purchasing;
 
-import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerFilter;
-import org.labkey.api.data.SimpleFilter;
 import org.labkey.api.data.TableInfo;
-import org.labkey.api.query.FieldKey;
 import org.labkey.api.data.DatabaseTableType;
 import org.labkey.api.query.QueryUpdateService;
 import org.labkey.api.query.SimpleUserSchema;
 import org.labkey.api.query.SimpleQueryUpdateService;
-import org.labkey.api.security.User;
-import org.labkey.api.security.permissions.UpdatePermission;
 
 public class EHR_PurchasingTable extends SimpleUserSchema.SimpleTable<EHR_PurchasingUserSchema>
 {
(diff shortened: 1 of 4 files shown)


labkey-jeckels (changes requested):
A few questions from my initial review:
1. Is the QueryUpdateService.updateRows() that takes a BatchValidationException the preferred one to call now? If so, can you add a comment to this effect and/or deprecate the one that doesn't take that argument?
2. Related to the first question, it seems clunky if we have to update all callers to effectively duplicate the code around creating the BatchValidationException and checking if it has errors and then throwing. Is this really an improvement to the API? Or is there some larger refactor that we should take to callers of updateRows() instead (though perhaps not right now)?
3. A comment in EHR_PurchasingTable about why it's necessary would be nice.

XingY:
> A few questions from my initial review:
> 
> 1. Is the QueryUpdateService.updateRows() that takes a BatchValidationException the preferred one to call now? If so, can you add a comment to this effect and/or deprecate the one that doesn't take that argument?
> 2. Related to the first question, it seems clunky if we have to update all callers to effectively duplicate the code around creating the BatchValidationException and checking if it has errors and then throwing. Is this really an improvement to the API? Or is there some larger refactor that we should take to callers of updateRows() instead (though perhaps not right now)?
> 3. A comment in EHR_PurchasingTable about why it's necessary would be nice.

1. The one with BatchValidationException is the preferred one. The one without really should be removed. I kept it there so that custom modules won't break on the next upgrade. I'll add comments in platform/ on my other branch.
2. The change to require explicit BatchValidationException passed in is to be consistent with the existing insertRows api. It does seem possible to refactor both apis to do a self contained errors check/throw. 
3. comment added.

labkey-jeckels (approved):

PR #533, opened 2023-02-14
Populate EHR Lookups alternate manifest

labkey-martyp:
#### Rationale
Update loading ehr lookups API to allow passing in a different (test) manifest to separate setting up a project from scratch (test) vs updating lookups in an existing EHR project. Setup lookups manifest URL parameter on populate lookups for testing. Minor tweaks for testing.

#### Related Pull Requests
* https://github.com/LabKey/johnsHopkinsEHRModules/pull/51

#### Changes
* Update PopulateLookupsAction as described above.
* Pass manifest to PopulateLookupsAction in UI if defined as URL parameter
* Tweaks for testing and query validation

### DIFF
--- a/ehr/test/src/org/labkey/test/tests/ehr/AbstractEHRTest.java
+++ b/ehr/test/src/org/labkey/test/tests/ehr/AbstractEHRTest.java
@@ -401,8 +401,9 @@ abstract public class AbstractEHRTest extends BaseWebDriverTest implements Advan
 
         doExtraPreStudyImportSetup();
 
-        populateInitialData();
         defineQCStates();
+        populateInitialData();
+
         // Do this first to establish the QC states before importing
         importStudy();
         disableMiniProfiler();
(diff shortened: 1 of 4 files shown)


ehr/test/src/org/labkey/test/tests/ehr/AbstractEHRTest.java
```diff
@@ -401,8 +401,9 @@ protected void initCreatedProject() throws Exception
 
         doExtraPreStudyImportSetup();
 
-        populateInitialData();
         defineQCStates();
+        populateInitialData();
```
labkey-jeckels, line 406: I take it that the ordering is important here. If so, adding a comment to capture that would be nice.

labkey-jeckels (approved):

ehr/test/src/org/labkey/test/tests/ehr/AbstractEHRTest.java
labkey-martyp: yeah need QC states before populating data

PR #541, opened 2023-02-17
Add parents to animal record in core EHR

labkey-martyp:
#### Rationale
Make demographicsParents query part of core EHR and add it to the animal record in core EHR. Most centers have demographicsParents already overwritten and are adding it as Parents to the animal record. This adds a base version of demographicsParents and will add it to the animal record if not already done so by the center. Includes supplemental pedigree.  

#### Changes
* demographicsParents query
* Add Parents to Animal table in DefaultEHRCustomizer

### DIFF
--- a/ehr/resources/queries/study/demographicsParents.sql
+++ b/ehr/resources/queries/study/demographicsParents.sql
@@ -26,7 +26,7 @@ SELECT
         END as numParents
 FROM study.demographics d
 
-UNION
+UNION ALL
 
 SELECT
     s.Id,
--- a/ehr/resources/queries/study/demographicsParents.sql
+++ b/ehr/resources/queries/study/demographicsParents.sql
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2015 LabKey Corporation
+ *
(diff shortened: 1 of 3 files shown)


ehr/resources/queries/study/demographicsParents.sql
```diff
@@ -0,0 +1,43 @@
...
+    d.sire,
+    'Observed' as sireType,
+    CASE
+        WHEN d.dam IS NOT NULL AND d.sire IS NOT NULL THEN 2
+        WHEN d.dam IS NOT NULL OR d.sire IS NOT NULL THEN 1
+        ELSE 0
+        END as numParents
+FROM study.demographics d
+
+UNION
```
ankurjuneja: UNION ALL is faster if the result of UNION is unique

ankurjuneja (approved):

labkey-jeckels (approved):

ehr/resources/queries/study/demographicsParents.sql
labkey-martyp: Yeah good call. I guess we'll make that assumption that there won't be duplicate records in supplemental pedigree and demographics.

PR #543, opened 2023-02-20
EHR Lookups form updates and automatically close records 

labkey-martyp:
#### Rationale
Update EHR Lookups UI to not show checkboxes and show all lookups on one page. Use title instead of value and display description for more human readable and easier filtering.

Add trigger property listing datasets that should automatically close out any previous records before entering a new record. 

#### Related Pull Requests
* https://github.com/LabKey/johnsHopkinsEHRModules/pull/57

#### Changes
* EHR Lookups: opt out of selection and show up to 500 values before pagination.
* EHR Lookups: Add edit link to title column and update default view to show title, category and description
* Add datasetsToCloseOnNewEntry trigger property and helper function to close all previous records for a given animal and dataset. Uses existing Java trigger helper to close records.
* In complete trigger automatically close specified records in datasetsToCloseOnNewEntry
* Refactor housing trigger to use the same record closing function

### DIFF
--- a/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
+++ b/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
@@ -2412,7 +2412,12 @@ public class TriggerScriptHelper
 
     public void closeHousingRecords(List<Map<String, Object>> records) throws Exception
     {
-        TableInfo housing = getTableInfo("study", "housing");
+        closePreviousDatasetRecords("housing", records, true);
+    }
+
+    public void closePreviousDatasetRecords(String dataset, List<Map<String, Object>> records, boolean validateDateTime) throws Exception
+    {
+        TableInfo datasetTi = getTableInfo("study", dataset);
         List<Map<String, Object>> toUpdate = new ArrayList<>();
         List<Map<String, Object>> oldKeys = new ArrayList<>();
 
@@ -2444,10 +2449,10 @@ public class TriggerScriptHelper
         {
             Date date = _dateTimeFormat.parse(row.get("date").toString());
             // TODO how do we override this a center specific module can opt out of this check?
-            if (date.getHours() == 0 && date.getMinutes() == 0)
+            if (validateDateTime && date.getHours() == 0 && date.getMinutes() == 0)
             {
                 Exception e = new Exception();
-                _log.warn("Attempting to terminate housing records with a rounded date.  This might indicate upstream code is rounding the date: " + _dateTimeFormat.format(date), e);
+                _log.warn("Attempting to terminate " + dataset + " records with a rounded date.  This might indicate upstream code is rounding the date: " + _dateTimeFormat.format(date), e);
             }
 
             SimpleFilter filter = new SimpleFilter(FieldKey.fromString("Id"), row.get("Id"));
@@ -2461,7 +2466,7 @@ public class TriggerScriptHelper
                 filter.addCondition(FieldKey.fromString("lsid"), encounteredLsids, CompareType.NOT_IN);
             }
 
-            TableSelector ts = new TableSelector(housing, Collections.singleton("lsid"), filter, null);
+            TableSelector ts = new TableSelector(datasetTi, Collections.singleton("lsid"), filter, null);
             List<String> ret = ts.getArrayList(String.class);
             if (!ret.isEmpty())
             {
@@ -2482,10 +2487,10 @@ public class TriggerScriptHelper
 
         if (!toUpdate.isEmpty())
         {
-            _log.info("closing housing records: " + toUpdate.size());
+            _log.info("closing " + dataset + " records: " + toUpdate.size());
             Map<String, Object> context = getExtraContext();
             context.put("skipAnnounceChangedParticipants", true);
-            housing.getUpdateService().updateRows(getUser(), getContainer(), toUpdate, oldKeys, null, context);
+            datasetTi.getUpdateService().updateRows(getUser(), getContainer(), toUpdate, oldKeys, null, context);
         }
     }
 
--- a/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
+++ b/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
@@ -2412,10 +2412,10 @@ public class TriggerScriptHelper
 
     public void closeHousingRecords(List<Map<String, Object>> records) throws Exception
     {
-        closePreviousDatasetRecords("housing", records, true);
+        closePreviousDatasetRecords("housing", records, false, false);
     }
 
-    public void closePreviousDatasetRecords(String dataset, List<Map<String, Object>> records, boolean dateOnly) throws Exception
+    public void closePreviousDatasetRecords(String dataset, List<Map<String, Object>> records, boolean dateOnly, boolean publicData) throws Exception
     {
         TableInfo datasetTi = getTableInfo("study", dataset);
         List<Map<String, Object>> toUpdate = new ArrayList<>();
@@ -2460,6 +2460,10 @@ public class TriggerScriptHelper
             //we want to only close those records starting prior to this record
             filter.addCondition(FieldKey.fromString("date"), date, CompareType.LTE);
             filter.addCondition(FieldKey.fromString("objectid"), row.get("objectid"), CompareType.NEQ_OR_NULL);
+
+            if (publicData)
+                filter.addCondition(FieldKey.fromString("qcstate/publicdata"), true, CompareType.EQUAL);
+
             if (!encounteredLsids.isEmpty())
             {
                 filter.addCondition(FieldKey.fromString("lsid"), encounteredLsids, CompareType.NOT_IN);
--- a/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
+++ b/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
@@ -2415,7 +2415,7 @@ public class TriggerScriptHelper
         closePreviousDatasetRecords("housing", records, true);
     }
 
-    public void closePreviousDatasetRecords(String dataset, List<Map<String, Object>> records, boolean validateDateTime) throws Exception
+    public void closePreviousDatasetRecords(String dataset, List<Map<String, Object>> records, boolean dateOnly) throws Exception
     {
         TableInfo datasetTi = getTableInfo("study", dataset);
         List<Map<String, Object>> toUpdate = new ArrayList<>();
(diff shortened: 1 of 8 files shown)


bbimber:
@labkey-martyp: I'm not looking at the code for this comment, but how's this feature differ from the existing functionality on a whole lot of datasets to close out previous records (housing, assignment, etc)? Further, it doesnt seem like it considers QC state? Even if that's not a current customer use-case, I think it's a mistake not to treat QC=Complete (which might be the de facto thing happening here) differently from draft, etc..

labkey-martyp:
> @labkey-martyp: I'm not looking at the code for this comment, but how's this feature differ from the existing functionality on a whole lot of datasets to close out previous records (housing, assignment, etc)? Further, it doesnt seem like it considers QC state? Even if that's not a current customer use-case, I think it's a mistake not to treat QC=Complete (which might be the de facto thing happening here) differently from draft, etc..

Good call on qcstate. Updated to only do this on qcstate completed. Housing does similar it is just a little more built out to handle batches and auditing. I ended up refactoring closeHousingRecords to reuse that functionality for this case.

bbimber:
@labkey-martyp OK. Still two questions:

- What exactly is different about what you're trying to add from existing hooks?
- Are you familiar with the onBecomePublic hook? The idea of this is to capture the first insert/update in which a record becomes "public" (typically meaning QCState = Completed, but technically an instance could have other QCState labels)

labkey-martyp:
> @labkey-martyp OK. Still two questions:
> 
> * What exactly is different about what you're trying to add from existing hooks?
> * Are you familiar with the onBecomePublic hook? The idea of this is to capture the first insert/update in which a record becomes "public" (typically meaning QCState = Completed, but technically an instance could have other QCState labels)

This is a simple way to apply this behavior to multiple datasets without having to create hooks or repeating the same code for each dataset. It's applied just like the datasetsToClose property.

bbimber:
I see a little better now, and it's reasonable to generalize the housing code to other datasets, but I'm not sure this is going to work robustly. Primarily, why dont you just copy the complete pattern used in housing: https://github.com/LabKey/ehrModules/blob/0de6def6075d68e146893ab7b32735a426e05534/ehr/resources/queries/study/housing.js#L117

instead of half doing it here: https://github.com/LabKey/ehrModules/blob/994e38de3c91dc95d71a29bed055c5d33ba54998/ehr/resources/scripts/ehr/triggers.js#L184

I think there's a few reasons it is expecting to call this cleanup code at the end (where it acts on all records in the transaction, and is called after the transection), rather than per-record (and before the transaction is completed). You'll see the existing java code is sorting records and expect to check them in chronological order. This allows the enddate that is applied to be the most appropriate one. While I'll grant it's sort of a fringe case to get two housing records per ID in a transaction, there are valid cases for this and it'd give the wrong outcome depending on the sort order of the input records.

LabKey has changed it's behavior as different points in time, but at least when I wrote this stuff it was very important to call certain code either during the before* triggers or after* triggers. If memory serves, the records returned by the selects would either include the current transaction or not, and in this code I think we want it to include those records. Again, the query engine might have changed in 10 years, but I'd be concerned this code would miss some probably rare but valid scenarios.

This also strikes me as something where it is really important to ensure integration test cases are adding to validate expected behavior. While it's annoying that EHR requires so much context and setup to work, the ONPRC EHR tests have a decent example of using the java client API to execute various patterns of inserts and updates to rapidly check these behaviors.

labkey-martyp:
> I see a little better now, and it's reasonable to generalize the housing code to other datasets, but I'm not sure this is going to work robustly. Primarily, why dont you just copy the complete pattern used in housing:
> 
> https://github.com/LabKey/ehrModules/blob/0de6def6075d68e146893ab7b32735a426e05534/ehr/resources/queries/study/housing.js#L117
> 
> instead of half doing it here:
> 
> https://github.com/LabKey/ehrModules/blob/994e38de3c91dc95d71a29bed055c5d33ba54998/ehr/resources/scripts/ehr/triggers.js#L184
> 
> I think there's a few reasons it is expecting to call this cleanup code at the end (where it acts on all records in the transaction, and is called after the transection), rather than per-record (and before the transaction is completed). You'll see the existing java code is sorting records and expect to check them in chronological order. This allows the enddate that is applied to be the most appropriate one. While I'll grant it's sort of a fringe case to get two housing records per ID in a transaction, there are valid cases for this and it'd give the wrong outcome depending on the sort order of the input records.
> 
> LabKey has changed it's behavior as different points in time, but at least when I wrote this stuff it was very important to call certain code either during the before* triggers or after* triggers. If memory serves, the records returned by the selects would either include the current transaction or not, and in this code I think we want it to include those records. Again, the query engine might have changed in 10 years, but I'd be concerned this code would miss some probably rare but valid scenarios.
> 
> This also strikes me as something where it is really important to ensure integration test cases are adding to validate expected behavior. While it's annoying that EHR requires so much context and setup to work, the ONPRC EHR tests have a decent example of using the java client API to execute various patterns of inserts and updates to rapidly check these behaviors.

I agree it would be good to have the same consistent pattern for closing records. That's a little bigger refactor than we have budget for on this right now. Closing out records by date in the "complete" trigger doesn't work great if only using dates and not datetimes, as this trigger gets called again after the initial records are closed and closes out new records. Performing it at the row level doesn't fire the trigger again for the closed out record. 

Pretty much the entire saveRows call is in a single transaction so there shouldn't be any issues rolling back. 

We've started adding tests on some of this in the JHU module but I'll check out ONPRC for how we might test this.

I can file a ticket for the bigger refactor to further combine this and the housing records being closed out. This feature is opt-in behavior so it shouldn't affect anyone else.

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -2482,10 +2487,10 @@ public int compare(Map<String, Object> o1, Map<String, Object> o2)
 
         if (!toUpdate.isEmpty())
         {
-            _log.info("closing housing records: " + toUpdate.size());
+            _log.info("closing " + dataset + " records: " + toUpdate.size());
```
labkey-jeckels, line 2493: This seems like it could be `debug` level

labkey-jeckels (approved):

bbimber:
I get the point on scope, but implementing a half-complete pattern doesnt make sense to me either. Can you elaborate on what's stopping this from simply doing exactly what housing does now (i.e. iterate the records during complete())? It seems like you could pull that out extremely close to as-is, and just parameterize the table name? 

You already made closeHousingRecords() more generic. That function is designed to accept all records from this transaction, sort them, and do the bookkeeping in-order. If the trigger is getting re-called, it seems like we ought to just address that. Does this re-calling behavior happen for housing closures? This kind of circular behavior is precisely 

The reason transactions are potentially very important is not rollbacks, but the behavior of the TableSelectors in this method. If memory serves, and I will grant that it's possible LK's query behavior has changed, it used to be important to issue SELECTs after the SaveRows transaction was committed, and not during the before phase as you're doing here. 

If you have a problem with date-only records stomping over each other, it seems like the closePreviousDatasetRecords() ought to have an argument for 'dateOnly' (as opposed to just validateDateTime). If it's dateOnly, you skip that validation step, and you might also want a different filter type here: 

filter.addCondition(FieldKey.fromString("date"), date, CompareType.LTE);

What's the desired behavior? Is it allowable to have two records that start on the same day? in the dateOnly case, the filter type probably should be CompareType.DATE_LESS_THAN?

bbimber:
BTW - would bet that my point on transactions and issuing select during before or complete could explain why you're not seeing that recursive close-out when you call this from 'before', but you say you do when you call it from complete. I'd argue you actually want to include the current incoming records in this filter, since it is possible for the incoming records to have conflicts that should be getting closed.

labkey-martyp:
> BTW - would bet that my point on transactions and issuing select during before or complete could explain why you're not seeing that recursive close-out when you call this from 'before', but you say you do when you call it from complete. I'd argue you actually want to include the current incoming records in this filter, since it is possible for the incoming records to have conflicts that should be getting closed.

The conflicting behavior with date only records and closing out housing records in the complete trigger is when you have two records with the same date,
1. Complete trigger fires and closes all records whose date is LTE the new record, except the new one itself.
2. Closing out those records starts the dataiterator again and the complete trigger fires again with the closed records as the newly updated rows. They close out all records that are LTE which includes the initial record that was just added. So all records including the just added record are closed out.

We've gotten around this in custom module housing triggers by concatenating a time later in the day to the original record date, but we can't do that in core EHR.

With row level triggers, the row triggers won't fire again when updates are done from within the trigger, so the current implementation doesn't loop back and close out the current record.

bbimber:
comments to that. I know I'm being very picky on this, but I think this is ignoring prior lessons and possibly trying to shoehorn something into this that either isnt complete adjusted or isnt a fit:

- Maybe another center works differently, but the point with housing is that when using date/time, it's actually not allowable to have two records with the same date/time. Unless I'm missing it, the existing housing-centric code should be passing through a date/time, and therefore the condition where you have dual identical date/time rows shouldnt ever occur. The code already checks for objectId being equal. A lot of the housing code is based on the assumption of it being impossible to overlap.

- I can see how this code could work sub-optimally with a date-only record. It's not really designed for that. But I think some minor changes ought to more directly address that issue, rather than indirectly doing this like I believe this approach is. 

- The question I do not have a handle on is "what is the actual desired behavior when there's two record with the identical date-only date"? What kind of data has this situation? I'd like to know more about the context here. Like I said in my post above, I think having a much more explicit switch in behavior between comparisons designed to act in date-time vs. date-only feels like the right fix here. I think that ought to be a really discrete change in closePreviousDatasetRecords. As-is, the new validateDateTime argument is sort of a half-measure in that direction. 

- Which center is doing this thing where they append a newer time to the record, and is this visible? As noted above, if anywhere is not actually treating housing as date/time, i think that's a problem. I could nonetheless see a pretty easy way to allow data entry to be date-only, and have code (upstream of what we're talking about here), figure out the correct time to assign to the incoming record. 

Again, I would bet this code as-is has a couple blind spots:

- If you have multiple incoming rows for the same Id, especially if they are inserted with random sort order, I'm not sure this is going to do the right thing. The reason this is being called at the end during complete() is specifically so we can look at all incoming records together. As noted above, I also am pretty sure the TableSelects, when called before the commit, return an incomplete set of records. There is a very specific reason housing does this bookkeeping at the end. I'll grant this is a less common case, but a lot of EHR's code is build to ensure fringe cases work correctly.

labkey-martyp:
Thanks for the feedback on this Ben. I would argue that using the designed functionality of row level triggers to handle this case is not a shoehorn solution. And yes there are corner cases, particularly if entering non-real use case data all at the same time in the form, that have gaps in both implementations. 

That being said, to move forward on this I refactored the original close housing helper function to better account for dateOnly dates and QCState, and fixed the issue with looping back through multiple times in the triggers. That made it possible to cover the main use cases for this new feature and leaves the original EHR housing close-out behavior as-is. This is now called from the EHR complete trigger. The housing close out is still in the housing trigger because many of the centers have already overridden it so I don't want to force it somewhere else.

We're continuing to add more features like this going forward to make the EHR more broadly usable outside ONPRC's case. I appreciate you advocating for them and providing feedback. There will likely continue to be varying scenarios going forward.

bbimber:
Thanks Marty. I don't think the row-level was inherently a bad idea at all, but it goes backwards from a specific change learned after real-world use of this code path, fringe case or not. 

As we arrived at late in this thread, I actually think one of the main issues here is actually the desired behavior when comparing date vs. date-time. As you might have seen I made a test case here (https://github.com/LabKey/ehrModules/pull/547). Having a test case this might have informed these questions on what is or isnt working as desired. You're also right that the current code (at least before this PR, but I dont think this would change it) doesnt handle records with identical date/time well. At least in the case of housing, having identical starts is probably not a valid condition, so it probably should fail upstream, but your point is right there.

PR #545, opened 2023-02-21
Ticket 46819: Refactor ClinicalEncountersStore

labkey-martyp:
#### Rationale
Currently trying to use the ClinicalEncountersClientStore before it is loaded causing logs filled with errors.

#### Changes
* If the store is not loaded skip trying to get records

### DIFF
--- a/ehr/resources/web/ehr/data/ClinicalEncountersClientStore.js
+++ b/ehr/resources/web/ehr/data/ClinicalEncountersClientStore.js
@@ -24,19 +24,22 @@ Ext4.define('EHR.data.ClinicalEncountersClientStore', {
         this.callParent(arguments);
     },
 
-    onRecordUpdate: function(record, modifiedFieldNames){
-        if (record.get('procedureid')){
-            modifiedFieldNames = modifiedFieldNames || [];
+    onRecordUpdate: function(record, modifiedFieldNames) {
+        if (record.get('procedureid')) {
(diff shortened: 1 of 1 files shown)


ehr/resources/web/ehr/data/ClinicalEncountersClientStore.js
```diff
@@ -97,15 +100,26 @@ Ext4.define('EHR.data.ClinicalEncountersClientStore', {
...
         var procRecIdx = procedureStore.findExact('rowid', procedureId);
         LDK.Assert.assertTrue('Unable to find procedure record in ClinicalEncountersClientStore for procedureId: [' + procedureId + ']', procRecIdx > -1);
 
         var procedureRec = procedureStore.getAt(procRecIdx);
         LDK.Assert.assertNotEmpty('Unable to find procedure record from ClinicalEncountersClientStore.  ProcedureId was: [' + procedureId + ']', procedureRec);
 
         return procedureRec;
+    },
+
+    getProcedureRecord: function(procedureId, modifiedFieldNames, cb){
```
labkey-jeckels: A little doc here on the expected arguments for the callback function would be nice, and a note that it may be invoked synchronously or async

labkey-jeckels (approved):

ehr/resources/web/ehr/data/ClinicalEncountersClientStore.js
labkey-martyp: added a comment

PR #549, opened 2023-02-22
Make module specific animal history reports override core EHR reports

labkey-martyp:
#### Rationale
This is the next iteration on the new paradigm of maintaining a core set of animal history reports and adding additional reports in the individual PRC modules. This PR allows overriding the core reports with those defined in additionalReports based on the report name.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4139

#### Changes
* Update PopulateReportsAction to merge in additionalReports (based on reportname) with the replace option

### DIFF
--- a/ehr/src/org/labkey/ehr/EHRController.java
+++ b/ehr/src/org/labkey/ehr/EHRController.java
@@ -1599,7 +1599,7 @@ public class EHRController extends SpringActionController
                         context = new DataIteratorContext(batchErrors);
                         context.setConfigParameters(configParameters);
                         context.setInsertOption(QueryUpdateService.InsertOption.REPLACE);
-                        context.setAlternateKeys(Set.of("reportname"));
+                        context.getAlternateKeys().add("reportname");
 
                         loader = DataLoader.get().createLoader(_additionalReportsResource, true, null, TabLoader.TSV_FILE_TYPE);
                         AbstractQueryImportAction.importData(loader, table, updateService, context, auditEvent, getUser(), getContainer());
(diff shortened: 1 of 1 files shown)


ehr/src/org/labkey/ehr/EHRController.java
```diff
@@ -1582,13 +1591,18 @@ public Object execute(PopulateLookupsForm form, BindException errors) throws Exc
...
                     {
+                        // Merge in additional reports (based on report name) with replace option
                         responseText.append("Populating additional reports from " + _additionalReportsModule.getName() + " module\n");
                         if (behaviorType != null && behaviorType != AuditBehaviorType.NONE)
                             auditEvent = createTransactionAuditEvent(getContainer(), QueryService.AuditAction.INSERT);
 
+                        context = new DataIteratorContext(batchErrors);
+                        context.setConfigParameters(configParameters);
+                        context.setInsertOption(QueryUpdateService.InsertOption.REPLACE);
+                        context.setAlternateKeys(Set.of("reportname"));
```
labkey-jeckels: Does this work as desired in dev machines that may have multiple EHR containers with reports loaded? There may be multiple existing rows with the same `reportname` - does this operate on just the ones in the current container? And should we have a constraint on `container, reportname`?

labkey-jeckels (commented):
One question before merging.

ehr/src/org/labkey/ehr/EHRController.java
labkey-martyp: Yes this respects the container, so it won't replace a report in another EHR container. 

I would like to add that constraint but it looks like at least ONPRC has multiple reports with the same name. They are not using this action to populate reports so it won't be an issue for them, but they are using the ehr.reports table so the constraint would have a conflict if we tried to add it to the table. I can work with them to rename those reports to unique names but that may take a while to get done.

labkey-jeckels (approved):

PR #574, opened 2023-03-30
When adding QueryForeignKey, ensure target container is properly set

bbimber:
I'm pretty sure this code has been subtly wrong for a long time. The purpose of the code in DefaultEHRCustomizer is to set a FK on columns where conceptURI is a participant, pointing to study.animal. It currently adds this FK without setting the target container. That works perfectly if operating within one container. 

As-is, the code calls getEHRStudyUserSchema() before adding this lookup, which will find the configured EHR source folder. That's supposed to make this work across folders. However, when DefaultEHRCustomizer adds the QueryForeignKey, it currently leaves container null. This will end up defaulting to the current container, which isnt going to be correct in the cross-folder scenario. 

Update: much of the disconnect is related to containerFilter. The default container filter uses the container of the source schema, not the target schema. This is addressed in the associated PR on platform/4307 . 

The change here is setting ContainerFilter to NULL, which is a reasonable thing to do and probably necessary for this to work properly across containers, but this alone doesnt address the disconnect happening in AbstractForeignKey. 

For future readers: this was edited after the discussion below.

### DIFF
--- a/ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
+++ b/ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
@@ -235,7 +235,7 @@ public class DefaultEHRCustomizer extends AbstractTableCustomizer
                     UserSchema us = getEHRStudyUserSchema(ti);
                     if (us != null)
                     {
-                        col.setFk(new QueryForeignKey(ti.getUserSchema(), ti.getContainerFilter(), us, null, "Animal", "Id", "Id"));
+                        col.setFk(new QueryForeignKey(ti.getUserSchema(), ti.getContainerFilter(), us, us.getContainer(), "Animal", "Id", "Id"));
                         col.setURL(DetailsURL.fromString("/ehr/participantView.view?participantId=${" + col.getName() + "}", us.getContainer()));
                     }
                 }
--- a/ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
+++ b/ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
@@ -235,7 +235,7 @@ public class DefaultEHRCustomizer extends AbstractTableCustomizer
                     UserSchema us = getEHRStudyUserSchema(ti);
                     if (us != null)
                     {
-                        col.setFk(new QueryForeignKey(ti.getUserSchema(), ti.getContainerFilter(), us, us.getContainer(), "Animal", "Id", "Id"));
+                        col.setFk(new QueryForeignKey(ti.getUserSchema(), null, us, null, "Animal", "Id", "Id"));
                         col.setURL(DetailsURL.fromString("/ehr/participantView.view?participantId=${" + col.getName() + "}", us.getContainer()));
                     }
                 }
@@ -257,7 +257,7 @@ public class DefaultEHRCustomizer extends AbstractTableCustomizer
                     {
                         UserSchema study = getEHRStudyUserSchema(ti);
                         if (study != null)
-                            col.setFk(new QueryForeignKey(ti.getUserSchema(), ti.getContainerFilter(), study, null, "Clinpath Runs", "objectid", ID_COL));
+                            col.setFk(new QueryForeignKey(ti.getUserSchema(), null, study, null, "Clinpath Runs", "objectid", ID_COL));
                     }
                     col.setFacetingBehaviorType(FacetingBehaviorType.ALWAYS_OFF);
                     col.setUserEditable(false);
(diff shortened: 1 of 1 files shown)


labkey-martyp:
Thanks @bbimber that change makes sense. Just waiting on TC to run the tests then we'll get this merged in.

ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
```diff
@@ -235,7 +235,7 @@ private void doSharedCustomization(AbstractTableInfo ti)
                     UserSchema us = getEHRStudyUserSchema(ti);
                     if (us != null)
                     {
-                        col.setFk(new QueryForeignKey(ti.getUserSchema(), ti.getContainerFilter(), us, null, "Animal", "Id", "Id"));
+                        col.setFk(new QueryForeignKey(ti.getUserSchema(), ti.getContainerFilter(), us, us.getContainer(), "Animal", "Id", "Id"));
```
labkey-jeckels: I understand the concern here, but in my reading of the code and some quick testing, when `lookupContainer` is null, we use the container from the `UserSchema` that's the third argument to the constructor here. So this change is harmless but already matches the existing behavior.

bbimber: That's exactly how I was expecting this to behave, but it was not doing this in practice.  I can try to work up a local repro, but this is an instance of needing a lot of setup/context to get it to behave right.

labkey-jeckels: OK. I didn't explore the second argument - presumably it's using that ContainerFilter when it generates SQL (pointing at container A) even though the TableInfo thinks that it's bound to the container B. If you're seeing a real-world problem somewhere that's incentive enough to take the change. It wasn't clear to me from your description if this was hypothetical or not.

bbimber: I'll try to find next week to dust off my local EHR instance. If I get an EHR running it should be relatively easy to repro this. I'll try to get a definitive answer on where it's breaking down.

labkey-jeckels: I didn't do a full repro, but I was able to verify that the ContainerFilter being passed in is indeed putting the current container ID's into the generated SQL. That would explain what you're seeing. I think this is more readable and also avoids passing the ContainerFilter.

```suggestion
                        col.setFk(new QueryForeignKey.Builder(us, null).table("Animal").key("Id").display("Id"));
```

There are a bunch of other places that are following the same pattern for other lookups. Many will only be used in the context of the EHR container, but if this suggestion works for you, we should do a wider review.

bbimber: OK, you could well be right. I unfortunately am under some deadlines and spend way too much time in dataintegration this week, but I will get this running locally to actually inspect it. I would believe this isnt the only place we should consider changing.

bbimber:
@labkey-jeckels: a little more information. I partially have a local repo going. 

For the sake of having more background on the thread:

The background is that in ancillary one-off tables on our servers (like a random list in some lab's folder), it is often useful to register some common behaviors, including a lookup to EHR when this table holds an Animal column. It does this by registering EHR's WrappingTableCustomizer. This looks for PARTICIPANT_CONCEPT_URI (which is overloaded and is also used by other LabKey code). The issue I was hitting is that the lookup was getting added, but failed to resolve any values. I was assuming this was because it was generating that lookup to the current container, not the container holding the study.

The XML being added is something like this:

```
<tables xmlns="http://labkey.org/data/xml">
  <table tableName="SerumChemistry" tableDbType="NOT_IN_DB">
    <javaCustomizer class="org.labkey.ehr.table.WrappingTableCustomizer" />
    <tableTitle>Serum Chemistry</tableTitle>
    <columns>
      <column columnName="AnimalId">
        <conceptURI>http://cpas.labkey.com/Study#ParticipantId</conceptURI>        
      </column>
    </columns>
  </table>
</tables>
```

You are right that there are multiple arguable issues here, and I think there are some general examples where cross-folder FKs dont do the right thing. For example:

- The solution proposed here of explicitly setting target container does fix the problem.
- If we dont set container on the QueryForeignKey, you are correct that part of the problem comes from a problematic ContainerFilter. Grabbing the source table's container filter is probably also wrong, and this containerFilter is typically CurrentContainer, where the container is set to the source table (the list in this case), and it operates wrong.

So the fix here does mitigate EHR's problems.

There's some questionable code in AbstractForeignKey too though:

- AbstractForeignKey.getLookupContainerFilter() is called when the FK is made. If there is a null lookupContainer and a null containerFilter, line 121 grabs the sourceSchema and uses its defaultContainerFilter(), which is based on the current container and not the target of the FK. I think a simple fix would be to implement getSchema() in AbstractForeignKey. This ensures that the most appropriate container is being used, not just the current one. The patch below does this. 


[QFK_Container.patch](https://github.com/LabKey/ehrModules/files/11124045/QFK_Container.patch)

bbimber:
Actually, this is a better patch. Renaming getSchema() to getLookupSchema() makes the change marginally bigger, but it seemed like a good idea to separate getSourceSchema() and getLookupSchema().

[QFK_Container.patch](https://github.com/LabKey/ehrModules/files/11124194/QFK_Container.patch)

bbimber:
@labkey-jeckels or others: see PR here that adds the change to AbstractForeignKey's ContainerFilter behavior, which I think was part of the problem here: https://github.com/LabKey/platform/pull/4276

Per usual, I cant push to labkey/platform, so would it be possible for someone to copy bbimber/23.3_fb_QueryForeignKey?

labkey-jeckels (approved):

PR #577, opened 2023-04-13
Handle Bulk Edit even if there are more than one values in the selected rows

labkey-bpatel:
#### Rationale
Bulk Edit on WNPRC Billing's Data Entry form 'Enter Charges without Animal Ids ' crashes when the selected rows with Date field has more than one date, and this is because the Date field also has the time component to it, which can be different if the rows were selected at different times; and Bulk Edit expects it to be the same value across selected rows.

#### Changes
* Set default value even if there is more than one value in the selected rows for a field.

### DIFF
--- a/ehr/resources/web/ehr/panel/BulkEditPanel.js
+++ b/ehr/resources/web/ehr/panel/BulkEditPanel.js
@@ -29,9 +29,6 @@ Ext4.define('EHR.panel.BulkEditPanel', {
     title: 'Bulk Edit',
 
     initComponent: function(){
-
-        LDK.Assert.assertTrue('formConfig has no fields', this.formConfig.fieldConfigs.length > 0);
-
         Ext4.apply(this, {
             itemId: 'formPanel',
             border: false,
@@ -113,7 +110,6 @@ Ext4.define('EHR.panel.BulkEditPanel', {
     getFieldConfigs: function(){
         var fields = this.callParent(arguments);
         var newItems = [];
-        LDK.Assert.assertTrue('Fields not found for bulk editing', fields.length > 0);
         Ext4.Array.forEach(fields, function(item){
             item.originalDisabled = item.disabled;
             item.disabled = true;
--- a/ehr/resources/web/ehr/panel/BulkEditPanel.js
+++ b/ehr/resources/web/ehr/panel/BulkEditPanel.js
@@ -29,6 +29,9 @@ Ext4.define('EHR.panel.BulkEditPanel', {
     title: 'Bulk Edit',
 
     initComponent: function(){
+
+        LDK.Assert.assertTrue('formConfig has no fields', this.formConfig.fieldConfigs.length > 0);
+
         Ext4.apply(this, {
             itemId: 'formPanel',
             border: false,
@@ -80,7 +83,10 @@ Ext4.define('EHR.panel.BulkEditPanel', {
                 }, this);
 
                 var keys = Ext4.Object.getKeys(values);
-                if (keys.length == 1){
+
+                // there can be more than one value for certain fields, ex. Date field with the time component,
+                // there might be different times in the selected rows, so pick one as a default value
+                if (keys.length >= 1){
                     valMap[field.name] = values[keys[0]];
                 }
             }, this);
@@ -107,6 +113,7 @@ Ext4.define('EHR.panel.BulkEditPanel', {
     getFieldConfigs: function(){
         var fields = this.callParent(arguments);
         var newItems = [];
+        LDK.Assert.assertTrue('Fields not found for bulk editing', fields.length > 0);
         Ext4.Array.forEach(fields, function(item){
             item.originalDisabled = item.disabled;
             item.disabled = true;
--- a/ehr/test/src/org/labkey/test/util/ehr/EHRTestHelper.java
+++ b/ehr/test/src/org/labkey/test/util/ehr/EHRTestHelper.java
@@ -230,7 +230,7 @@ public class EHRTestHelper
         grid.clickTbarButton(btnLabel);
         grid.waitForRowCount(count + 1);
         grid.cancelEdit();
-        _test.sleep(700);
+        _test.sleep(200);
     }
 
     public void clickExt4WindowBtn(String title, String label)
--- a/ehr/test/src/org/labkey/test/util/ehr/EHRTestHelper.java
+++ b/ehr/test/src/org/labkey/test/util/ehr/EHRTestHelper.java
@@ -230,7 +230,7 @@ public class EHRTestHelper
         grid.clickTbarButton(btnLabel);
         grid.waitForRowCount(count + 1);
         grid.cancelEdit();
-        _test.sleep(200);
+        _test.sleep(700);
     }
 
     public void clickExt4WindowBtn(String title, String label)


ehr/resources/web/ehr/panel/BulkEditPanel.js
```diff
@@ -80,7 +80,10 @@ Ext4.define('EHR.panel.BulkEditPanel', {
                 }, this);
 
                 var keys = Ext4.Object.getKeys(values);
-                if (keys.length == 1){
+
+                // there can be more than one value for certain fields, ex. Date field with the time component,
+                // there might be different times in the selected rows, so pick one as a default value
+                if (keys.length >= 1){
```
labkey-jeckels, line 86: This doesn't feel quite right but maybe I'm not following it correctly. It seems like it will populate the Bulk Edit dialog with a value for *any* field that has at least one value, even though they're not the same. So if you have a field where rows have values 2, 5, and 9, it'll just choose one of them to populate the dialog?

The intent of this code is to have the dialog pop up with prepopulated with any values that are identical across the rows, and leave the rest blank. Is there a problem when none of the fields have matching values across all of the rows?

bbimber: This is definitely not how this is designed to work. What is breaking for WNPRC?

labkey-bpatel: @labkey-jeckels 
"So if you have a field where rows have values 2, 5, and 9, it'll just choose one of them to populate the dialog?"
With my change, that is correct.

"Is there a problem when none of the fields have matching values across all of the rows?"
Yes, its the same issue where the Bulk Edit window collapses.

@bbimber This is an attempt to fix issue as seen here: https://teamcity.labkey.org/viewLog.html?buildId=2389394&buildTypeId=LabKey_227Release_Premium_Ehr_WnprcEhrPostgres

bbimber: @labkey-bpatel: OK. I get there is a JS error happening. When you repro locally, what's the actual exception? 

The way this component is supposed to work is that if you have 3 rows selected, it should find all instances where 100% of the rows have the same value. For any field where this is true, it should pre-populate the field. 

I could imagine there are maybe some type-conversion situations that could break it? It has definitely been used on situations involving date, date-time, and similar types of values before.

labkey-bpatel: If a field across selected rows have different values, then we can change it such that we do not allow Bulk Edit in the first place/only allow Bulk Edit when the selected rows have same values. Thoughts?

labkey-bpatel: @bbimber 
client side exception from chrome:
 at constructor.finishRender (ext-all-sandbox-debug.js?980774510:28219:42)
    at constructor.finishRenderItems (ext-all-sandbox-debug.js?980774510:35320:22)
    at constructor.finishRender (ext-all-sandbox-debug.js?980774510:35675:12)
    at constructor.finishRenderChildren (ext-all-sandbox-debug.js?980774510:37892:20)
    at constructor.afterRender (ext-all-sandbox-debug.js?980774510:27968:12)
    at constructor.callParent (ext-all-sandbox-debug.js?980774510:4272:32)
    at constructor.afterRender (ext-patches.js?980774510:136:16)
    at constructor.callParent (ext-all-sandbox-debug.js?980774510:4272:32)
    at constructor.afterRender (ext-all-sandbox-debug.js?980774510:33625:12)
    at constructor.finishRender (ext-all-sandbox-debug.js?980774510:28246:12)
finishRender @ ext-all-sandbox-debug.js?980774510:28219
finishRenderItems @ ext-all-sandbox-debug.js?980774510:35320
finishRender @ ext-all-sandbox-debug.js?980774510:35675
finishRenderChildren @ ext-all-sandbox-debug.js?980774510:37892
afterRender @ ext-all-sandbox-debug.js?980774510:27968
callParent @ ext-all-sandbox-debug.js?980774510:4272
afterRender @ ext-patches.js?980774510:136
callParent @ ext-all-sandbox-debug.js?980774510:4272
afterRender @ ext-all-sandbox-debug.js?980774510:33625
finishRender @ ext-all-sandbox-debug.js?980774510:28246
finishRenderItems @ ext-all-sandbox-debug.js?980774510:35320
finishRender @ ext-all-sandbox-debug.js?980774510:35675
finishRenderChildren @ ext-all-sandbox-debug.js?980774510:37892
afterRender @ ext-all-sandbox-debug.js?980774510:27968
callParent @ ext-all-sandbox-debug.js?980774510:4272
afterRender @ ext-patches.js?980774510:136
callParent @ ext-all-sandbox-debug.js?980774510:4272
afterRender @ ext-all-sandbox-debug.js?980774510:33625
finishRender @ ext-all-sandbox-debug.js?980774510:28246
finishRenderItems @ ext-all-sandbox-debug.js?980774510:35320
finishRender @ ext-all-sandbox-debug.js?980774510:35675
finishRenderChildren @ ext-all-sandbox-debug.js?980774510:37892
afterRender @ ext-all-sandbox-debug.js?980774510:27968
callParent @ ext-all-sandbox-debug.js?980774510:4272
afterRender @ ext-patches.js?980774510:136
callParent @ ext-all-sandbox-debug.js?980774510:4272
afterRender @ ext-all-sandbox-debug.js?980774510:33625
finishRender @ ext-all-sandbox-debug.js?980774510:28246
finishRenderItems @ ext-all-sandbox-debug.js?980774510:35320
finishRender @ ext-all-sandbox-debug.js?980774510:35675
finishRenderChildren @ ext-all-sandbox-debug.js?980774510:37892
afterRender @ ext-all-sandbox-debug.js?980774510:27968
callParent @ ext-all-sandbox-debug.js?980774510:4272
afterRender @ ext-patches.js?980774510:136
callParent @ ext-all-sandbox-debug.js?980774510:4272
afterRender @ ext-all-sandbox-debug.js?980774510:33625
callParent @ ext-all-sandbox-debug.js?980774510:4272
afterRender @ ext-all-sandbox-debug.js?980774510:85176
finishRender @ ext-all-sandbox-debug.js?980774510:28246
render @ ext-all-sandbox-debug.js?980774510:28512
doAutoRender @ ext-all-sandbox-debug.js?980774510:28118
show @ ext-all-sandbox-debug.js?980774510:34016
handler @ BulkEditWindow.js?980774510:96
callback @ ext-all-sandbox-debug.js?980774510:7250
onClick @ ext-all-sandbox-debug.js?980774510:123187
onClick @ ext-all-sandbox-debug.js?980774510:124053
eval @ VM1411:6
wrap @ ext-all-sandbox-debug.js?980774510:10572

Could be related to type conversion, but it looks like even if I use plain and simple values across rows (screenshot below), its the same issue:
image

bbimber: I believe that would defeat the purpose of this.

Do you have a local repro, and what specific js error is happening?

That behavior where an extjs window renders to the side often happens when an xtype isn't loaded

labkey-bpatel: Yeah seems like it.

Yes, I have a local repro. Above is the only js error I am seeing.

Hmm, if xtype isn't getting loaded then I am not seeing an error related to that, let me investigate what xtype isn't getting loaded - thanks for the insight.

bbimber: what about commenting out the beforerender listener? does it still break?

labkey-bpatel: Looks like it (all the fields across rows are blank)
image

labkey-bpatel: "The way this component is supposed to work is that if you have 3 rows selected, it should find all instances where 100% of the rows have the same value. For any field where this is true, it should pre-populate the field."

So, it is currently not working when all the fields are null across rows (100% of the rows with the same null values).

It also does not work in this case...
image

..unless, the Date field is populated - it looks like that is the requirement from how its currently behaving, and that Date field contains both date and time, which has to be the same across rows; but that is the problem, datetime is not always the same since time can be slightly different, hence it crashes.

So it crashes in multiple scenarios:
1. If all the fields are null
2. If a field across rows have different values
3. If the Date field is left blank and other fields are filled out with the same value

bbimber: I dont suppose LabKey has a demo instance of EHR anywhere that's viewable? I believe you are seeing what you're seeing, but I am skeptical of the root cause.  

Did you see my comment about removing the beforerender listener? The sort of behavior you're seeing with a window dying usually happens with a buried JS error somewhere in the code. For example, I believe the data combinations you point out are putting it into some state that throws an error, but I dont believe solutions like grabbing the first value is going to be the right fix.

Is this specific to one form type, or a more general thing?

bbimber: I'm interested in the render listener because of this kind of thing:
```
var cols = win.down('#formPanel').items.get(0).items.getCount();
```

Let's say there's something about this combination of inputs that left "items" empty. Then items.get(0) throws an exception, which could propagate weirdly through Ext4's rendering. I dont know that this is your problem, but this kind of thing is what I often find causing the mode of failure you have here

bbimber:
For posterity, I want to correct the description above. The expected behavior is:

1) Component inspects all selected rows. If all rows have an identical value for a field, use that. If they do not, leave value null
2) Render a FormPanel with one input per field. If all input records had the same value for this field, set the input with that value. Otherwise leave the field null. The intention of this behavior is that if all records had the same value, then "no change" leaves them with this value. 
3) Arbitrarily picking the first value to set the field when records differ breaks the intent of the component
4) The fields render disabled. The user can click any field to enable it, and set a value. 
5) On submit, edited fields are used to update the records. This means that any field not touched would be unchanged in source records. The point is to let a user bulk edit the fields they want, but not change fields they dont. 

Again, my best guess is that there is something about this combination of data that breaks some expectation in the component's code, and that exception is giving the render error you see. There are some possible race conditions here related to loading the form config (i.e. maybe something in the code is trying to find components in the form prior to when it loaded metadata and added those fields).

ehr/resources/web/ehr/panel/BulkEditPanel.js
labkey-bpatel: Yes, I looked at that as well, however, commenting out beforerenderer did not make any difference.

labkey-bpatel:
> For posterity, I want to correct the description above. The expected behavior is:
> 
> 1. Component inspects all selected rows. If all rows have an identical value for a field, use that. If they do not, leave value null
> 2. Render a FormPanel with one input per field. If all input records had the same value for this field, set the input with that value. Otherwise leave the field null. The intention of this behavior is that if all records had the same value, then "no change" leaves them with this value.
> 3. Arbitrarily picking the first value to set the field when records differ breaks the intent of the component
> 4. The fields render disabled. The user can click any field to enable it, and set a value.
> 5. On submit, edited fields are used to update the records. This means that any field not touched would be unchanged in source records. The point is to let a user bulk edit the fields they want, but not change fields they dont.
> 
> Again, my best guess is that there is something about this combination of data that breaks some expectation in the component's code, and that exception is giving the render error you see. There are some possible race conditions here related to loading the form config (i.e. maybe something in the code is trying to find components in the form prior to when it loaded metadata and added those fields).

Thank you for the explanation. What it looks like is that the date field in the form is the culprit, setting a default value for the Date field in the config solves the problem we are seeing: https://github.com/LabKey/wnprc-modules/pull/291
With this change, it works as you are explaining above, now what I am not clear about is why this date field is the culprit, but that is the observed behavior.

Closing this PR

bbimber:
@labkey-bpatel: here's what I think happens. This is the WNPRC field for charge items: 

https://github.com/LabKey/wnprc-modules/blob/e3d2154c3df00da9973e8b6695eae57dbe49b757/wnprc_billing/resources/web/wnprc_billing/form/field/ChargeItemField.js#L125

The field has code that runs on init and expects a date value to be non-null. Your form has a null date. The way Ext4 deals with exceptions when loading a window obscures the problem, but I think this is your cause.

You can also see this if you simply load this grid and try to enter any kind of values when date is null. The form doesnt behave right (irrespective of bulk editing).

If you modify ChargeItemField like the following (which I doubt is actually good behavior, but it does avoid the exception):

```
    returnFilter.push(LABKEY.Filter.create('departmentCode', chargeGroup, LABKEY.Filter.Types.EQUAL));
    if (dateValue) {
        returnFilter.push(LABKEY.Filter.create('startDate', dateValue.format("Y-m-d"), LABKEY.Filter.Types.DATE_LESS_THAN_OR_EQUAL));
        returnFilter.push(LABKEY.Filter.create('endDate', dateValue.format("Y-m-d"), LABKEY.Filter.Types.DATE_GREATER_THAN_OR_EQUAL));
    }

```
then bulk edit will load. I dont really know the intent of what's happening in ChargeItemField, but I would guess that maybe it should abort and not load anything if there's a null date.

labkey-bpatel:
Applied your changes to ChargeItemField, and it accepts null values. Thank you for your help in troubleshooting. Closing this PR since the change is in wnprc-modules PR.

PR #581, opened 2023-04-17
Issue 408: Support deleting of records within a request

labkey-jeckels:
#### Rationale
Users sometimes need to completely delete a record from within a request. We can support this when it hasn't yet been attached to a task

#### Changes
* Differentiate between a record that's scoped to just the current request form vs other scenarios (like task-based forms)
* Give better feedback when we aren't going to actually delete

### DIFF
--- a/ehr/resources/web/ehr/data/DataEntryClientStore.js
+++ b/ehr/resources/web/ehr/data/DataEntryClientStore.js
@@ -334,6 +334,10 @@ Ext4.define('EHR.data.DataEntryClientStore', {
     },
 
     safeRemove: function(records){
+
+        let undeletableCount = 0;
+        let undeletableIds = {};
+
         Ext4.Array.forEach(records, function(r){
             // First check if the record is attached to a request
             if (!r.phantom && r.get('requestid')) {
@@ -342,12 +346,14 @@ Ext4.define('EHR.data.DataEntryClientStore', {
                 if (this.storeCollection.getRequestId &&
                         this.storeCollection.getRequestId() === r.get('requestid')) {
                     if (r.get('taskid')) {
-                        Ext4.Msg.alert('Unable to delete', 'The record is already attached to a task. It will be detached from the task but not deleted when saved.');
+                        let id = r.get('id') || 'Blank';
+                        undeletableIds[id] = true;
+                        undeletableCount++;
                         r.isRemovedRequest = true;
                     }
                     else {
-                        // Do nothing special. This will be a true delete, because we're editing the request its request
-                        // form, and the row isn't yet attached to a task
+                        // Do nothing special. This will be a true delete, because we're editing the request via its
+                        // request form, and the row isn't yet attached to a task
                     }
                 }
                 else {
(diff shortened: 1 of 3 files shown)


ankurjuneja (approved):

ankurjuneja (changes requested):
this is not working for me

bbimber:
Hi @labkey-jeckels, could comments/questions:

1) Whatever the answers below, I believe this needs a associated automated test case to be considered feature-complete. ONPRC_EHRTest2.bloodRequestTest() could probably be augmented relatively easily to include this logic, even if it's a Tulane feature. The behavior is general EHR. Like we discussed separately, I really think a general TestEHR module should ultimately get created (implementing a basic EHR instance and centralizing a lot of test cases from center-specific modules); however, adding to the existing ONPRC EHR test would be lowest effort here. This test augmentation should probably test: a) make multiple request records, b) assign to a task and then delete from it (where it's re-queued), c) move one of those requests to a higher QCState, d) open the request as a regular request user, e) try to delete the record that is "Request: Approved" and show it's denied, f) try to delete the request that's still "Request: Pending" and verify it actually is deleted.

To this feature:

1) If the user tries to delete multiple records, couldnt they in theory get many redundant alerts? Should you manage a flag and only generate the alert once per batch?

2)  Might it actually make more sense to implement some of this logic in DefaultClientModel.canDelete()? This currently tests QCState permissions (which i think will cover many of my potential concerns, since QCStates will advance and at least at ONPRC there are granular permissions around those states so most users cannot improperly delete requests once approved.

2B) This does not need to be addressed here, but I do think there is a very good chance that if this UI is going to be used regularly for allowing record deletes, and if records are unavailable for deleting because they have been assigned to a task or because their QCState advanced to "Request: Approved" and the user no longer has delete permission, then I think the grid as-is does not show enough information for the user to understand what's happening and there's a high chance of getting people confused and fustrated.

2C) If people really want to edit existing submissions, have you considered shuttling them toward the read-version of a task, rather than the entry form? This is the view that shows the form sections as data regions? Managing records post-submission via a data region is really how everything was designed to work. Again, it's not crazy to make deletion work in a rational manner via this form (which is doesnt currently), but I also dont think it's really the best place to have people manage requests post-submission.

ankurjuneja (approved):

labkey-jeckels:
@bbimber Thanks for the thoughts. TNPRC has a good amount of coverage for request scenarios so I've added it there, though their repo isn't public (I'll ask again if they can approve making it public). https://github.com/LabKey/tnprc_ehr/pull/1133 I didn't add coverage for the permission level checks for who can edit records in certain QCStates since there's existing coverage.

1. Yes, it would have shown multiple dialogs. I added code to coalesce and list affected animal IDs to help users understand which record(s) were impacted.
2. (and 2B) I can see adding that in the future if this gets more usage at centers who have different permissions for the users doing it. For now, users without permission will get errors from the trigger scripts when they try to save.
2C. That's where they've been directed as a workaround for the form's behavior. I think they're equally reasonable places for users to try to do the delete, and they're at least closer to consistent behavior with this PR.

PR #600, opened 2023-05-15
EHR Reports - Set import option

labkey-martyp:
#### Rationale
EHR Reports uses the Replace option when inserting additionalReports. This option now needs to be enabled for this table.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3996

#### Changes
* Enable replace import option in table customizer

### DIFF
--- a/ehr/src/org/labkey/ehr/EHRController.java
+++ b/ehr/src/org/labkey/ehr/EHRController.java
@@ -1579,7 +1579,7 @@ public class EHRController extends SpringActionController
 
                     Map<Enum, Object> configParameters = new HashMap<>();
                     configParameters.put(DetailedAuditLogDataIterator.AuditConfigs.AuditBehavior, behaviorType);
-                    configParameters.put(QueryUpdateService.ConfigParameters.SkipInsertOptionValidation, Boolean.TRUE);
+
                     context.setConfigParameters(configParameters);
                     context.setInsertOption(QueryUpdateService.InsertOption.INSERT);
 
@@ -1598,6 +1598,8 @@ public class EHRController extends SpringActionController
                             auditEvent = createTransactionAuditEvent(getContainer(), QueryService.AuditAction.INSERT);
 
                         context = new DataIteratorContext(batchErrors);
+                        configParameters.put(QueryUpdateService.ConfigParameters.SkipInsertOptionValidation, Boolean.TRUE);
+
                         context.setConfigParameters(configParameters);
                         context.setInsertOption(QueryUpdateService.InsertOption.REPLACE);
                         context.getAlternateKeys().add("reportname");
--- a/ehr/src/org/labkey/ehr/EHRController.java
+++ b/ehr/src/org/labkey/ehr/EHRController.java
@@ -1598,8 +1598,6 @@ public class EHRController extends SpringActionController
                             auditEvent = createTransactionAuditEvent(getContainer(), QueryService.AuditAction.INSERT);
 
                         context = new DataIteratorContext(batchErrors);
-                        configParameters.put(QueryUpdateService.ConfigParameters.SkipInsertOptionValidation, Boolean.TRUE);
-
                         context.setConfigParameters(configParameters);
                         context.setInsertOption(QueryUpdateService.InsertOption.REPLACE);
                         context.getAlternateKeys().add("reportname");
--- a/ehr/src/org/labkey/ehr/EHRController.java
+++ b/ehr/src/org/labkey/ehr/EHRController.java
@@ -1579,6 +1579,7 @@ public class EHRController extends SpringActionController
 
                     Map<Enum, Object> configParameters = new HashMap<>();
                     configParameters.put(DetailedAuditLogDataIterator.AuditConfigs.AuditBehavior, behaviorType);
+                    configParameters.put(QueryUpdateService.ConfigParameters.SkipInsertOptionValidation, Boolean.TRUE);
                     context.setConfigParameters(configParameters);
                     context.setInsertOption(QueryUpdateService.InsertOption.INSERT);
 
--- a/ehr/src/org/labkey/ehr/query/EHRCustomPermissionsTable.java
+++ b/ehr/src/org/labkey/ehr/query/EHRCustomPermissionsTable.java
@@ -17,6 +17,10 @@ public class EHRCustomPermissionsTable<SchemaType extends UserSchema> extends Cu
     public EHRCustomPermissionsTable(SchemaType schema, TableInfo table, ContainerFilter cf)
     {
         super(schema, table, cf);
+
+        setAllowedInsertOption(QueryUpdateService.InsertOption.MERGE);
+        setAllowedInsertOption(QueryUpdateService.InsertOption.REPLACE);
+        setAllowedInsertOption(QueryUpdateService.InsertOption.UPSERT);
     }
 
     @Override
--- a/ehr/src/org/labkey/ehr/query/EHRCustomPermissionsTable.java
+++ b/ehr/src/org/labkey/ehr/query/EHRCustomPermissionsTable.java
@@ -17,10 +17,6 @@ public class EHRCustomPermissionsTable<SchemaType extends UserSchema> extends Cu
     public EHRCustomPermissionsTable(SchemaType schema, TableInfo table, ContainerFilter cf)
     {
         super(schema, table, cf);
-
-        setAllowedInsertOption(QueryUpdateService.InsertOption.MERGE);
-        setAllowedInsertOption(QueryUpdateService.InsertOption.REPLACE);
-        setAllowedInsertOption(QueryUpdateService.InsertOption.UPSERT);
     }
 
     @Override
--- a/ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
+++ b/ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
@@ -56,6 +56,7 @@ import org.labkey.api.query.QueryDefinition;
 import org.labkey.api.query.QueryException;
 import org.labkey.api.query.QueryForeignKey;
 import org.labkey.api.query.QueryService;
+import org.labkey.api.query.QueryUpdateService;
 import org.labkey.api.query.UserSchema;
 import org.labkey.api.security.User;
 import org.labkey.api.settings.LookAndFeelProperties;
@@ -141,6 +142,10 @@ public class DefaultEHRCustomizer extends AbstractTableCustomizer
         else if (matches(table, "ehr", "protocol"))
         {
             customizeProtocolTable((AbstractTableInfo) table);
+        }
+        else if (matches(table, "ehr", "reports"))
+        {
+            customizeReportsTable((AbstractTableInfo) table);
         }
         else if (matches(table, "ehr_lookups", "procedures"))
         {
@@ -1402,6 +1407,13 @@ public class DefaultEHRCustomizer extends AbstractTableCustomizer
             displayCol.setLabel("Short Name");
             table.addColumn(displayCol);
         }
+    }
+
+    private void customizeReportsTable(AbstractTableInfo table)
+    {
+        // Enabling this option for populating reports from additionalReports.tsv. This is set here while data iterator
+        // refinements are being made and may be removed in the future.
+        table.setAllowedInsertOption(QueryUpdateService.InsertOption.REPLACE);
     }
 
     private void customizeProtocolTable(AbstractTableInfo table)


bbimber:
@labkey-martyp hey, just generally curious on what's going on with this. It wasnt 100% obvious looking at the code in the associated PR. Do I understand correctly that LK now has a new pathway to insert data that is a merge (i.e. update if the record exsits, insert if not)? And the issue here is that we're using alternate fields as the actual keys in the EHR lookups tables?

labkey-martyp:
@bbimber Yeah this is a new way to populate reports and lookups that is designed to consolidate more of the EHR into a base EHR setup. Essentially the reports.tsv is in EHR and if a center wants to add, override or remove reports then they can define an additionalReports.tsv in their module to make those changes. This action is doing a REPLACE import when doing the overrides and needs this flag set now.

This feature is only in use by the two most recent EHR build-outs and really just designed as a step toward establishing that base EHR case that can hopefully be quickly adopted by new EHR build outs. This probably does not have any significant value to the more advanced build-outs.

bbimber:
OK, cool. Seems like a useful addition. 

Two questions:

- In this instance, the table's true PK and the effective PK (i.e. the value in the incoming record) are different, right? Do we need any special configuration either in the table's metadata or in the incoming request to tell LabKey how to do that merge correctly?

- What exactly does the term "SkipInsertOptionValidation" refer to? That name is not intuitive at all to me. Guessing by the code, is that telling LK to skip some validation that happens by default (maybe over the alternate key field names or something)?

labkey-martyp:
@bbimber Sorry didn't really answer your question. I believe the core change is around better clarity to the users when an inert, merge or replace is happening (and some other fixes too). @XingY can probably explain better. 

First section in the 23.3 release notes about the import change, https://www.labkey.org/wiki/Documentation/Archive/23.3/page.view?name=releasenotes233#Data%20Import

1. This particular part is for EHR reports and we are specifying reportname as the alternate key. https://github.com/LabKey/ehrModules/blob/d40a6757477957c5513f24dfe924a1d426e83f79/ehr/src/org/labkey/ehr/EHRController.java#L1605
2. We've added some asserts that require this flag when importing using REPLACE, MERGE or UPSERT. I'll let @XingY explain the rest of the details on that. I'm not entirely clear on the reason.
https://github.com/LabKey/platform/blob/6840f253584e9cf831a9ba77cef90aba6a59ada9/api/src/org/labkey/api/query/AbstractQueryUpdateService.java#L323

XingY:
> 2\. https://github.com/LabKey/platform/blob/6840f253584e9cf831a9ba77cef90aba6a59ada9/api/src/org/labkey/api/query/AbstractQueryUpdateService.java#L323

Hi Ben,

SkipInsertOptionValidation generally speaking should be avoided. We added additional check that the insertOption must be one that the target table supports. If EHR Reports table uses Replace, then I think the correct changes to make is:
1. mark the Reports table (I'm not familiar with EHR schema, is this a single table or a collection of tables) as support Replace, by calling `setAllowedInsertOption(QueryUpdateService.InsertOption.REPLACE);
2. instead of insert, do `context.setInsertOption(QueryUpdateService.InsertOption.REPLACE)` near the changes that you are making

Ideally SkipInsertOptionValidation shouldn't exist. We had to add it to support special scenarios such as ETL or folder import. For ETL, we had allowed user to specify import option that a table doesn't necessary have full support for, and we don't want to break that. For folder import, Merge is always used. For other one off scenarios, the correct approach is to fix up the table itself

XingY (commented):
See comment. Depending on the implementable detail of EHR, we might end up having to bypass. But please explore the other option of fixing up the table and calller first.

bbimber:
> See comment. Depending on the implementable detail of EHR, we might end up having to bypass. But please explore the other option of fixing up the table and calller first.

@XingY, thanks for the reply. @labkey-martyp, you might actually be the right person to answer that since you're doing this work. What is it about ehr.reports that requires this override? 

That table is a custom class EHRCustomPermissionsTable. Does that code need to update or implement something to be compatible with this new QUS pathway?

labkey-martyp:
@XingY Why are these insert options opt-in? What's the downside to just allowing all insert options (merge, replace and upsert)?

bbimber:
@labkey-martyp, @XingY: This does not answer Marty's question, but as context: on the whole I would expect/hope while EHR has a custom UserSchema and some custom TableInfo subclasses, those currently all inherit from the standard AbstractTableInfo / DefaultUserSchema classes. We should be inheriting LK's default query-layer behaviors, and if there's something we're not correctly inheriting (perhaps some new behavior around merge/replace), then it seems like we ought ask the question why not.

XingY:
> @XingY Why are these insert options opt-in? What's the downside to just allowing all insert options (merge, replace and upsert)?

The reason for opt-in is scoping. To support insert options, often times custom code is needed, but it's impossible to determine that for every single table. If you know a table does support all insert options (because it's a very basic simple table?) then it's safe to opt-in.

bbimber:
> > @XingY Why are these insert options opt-in? What's the downside to just allowing all insert options (merge, replace and upsert)?
> 
> The reason for opt-in is scoping. To support insert options, often times custom code is needed, but it's impossible to determine that for every single table. If you know a table does support all insert options (because it's a very basic simple table?) then it's safe to opt-in.

@XingY: if this does not require customization for basic tables, such as module-provided tables with straightforward key fields, would it make sense to set these values in SimpleUserSchema.SimpleTable or something to that effect? This would let large categories of tables inherit that behavior, if it truly is that straightforward for that category of table. If there are restrictions around PKs, perhaps that logic could be implemented in SimpleTable, such as only enabling this if the table has a single PK (or whatever the restriction needs to be...).

XingY:
> > > @XingY Why are these insert options opt-in? What's the downside to just allowing all insert options (merge, replace and upsert)?
> > 
> > 
> > The reason for opt-in is scoping. To support insert options, often times custom code is needed, but it's impossible to determine that for every single table. If you know a table does support all insert options (because it's a very basic simple table?) then it's safe to opt-in.
> 
> @XingY: if this does not require customization for basic tables, such as module-provided tables with straightforward key fields, would it make sense to set these values in SimpleUserSchema.SimpleTable or something to that effect? This would let large categories of tables inherit that behavior, if it truly is that straightforward for that category of table. If there are restrictions around PKs, perhaps that logic could be implemented in SimpleTable, such as only enabling this if the table has a single PK (or whatever the restriction needs to be...).

The problem is almost every table extends SimpleTable. 

Here is the relevant PR and you will find more discussion around this. (Matt said in a comment" I think SimpleTable supports MERGE by default (this is sometimes a problem).
We could leave it this way by default, and try to track down the cases where we actually want MERGE to work, but there might be a lot of usages.")

https://github.com/LabKey/platform/pull/3949

bbimber:
> > > > @XingY Why are these insert options opt-in? What's the downside to just allowing all insert options (merge, replace and upsert)?
> > > 
> > > 
> > > The reason for opt-in is scoping. To support insert options, often times custom code is needed, but it's impossible to determine that for every single table. If you know a table does support all insert options (because it's a very basic simple table?) then it's safe to opt-in.
> > 
> > 
> > @XingY: if this does not require customization for basic tables, such as module-provided tables with straightforward key fields, would it make sense to set these values in SimpleUserSchema.SimpleTable or something to that effect? This would let large categories of tables inherit that behavior, if it truly is that straightforward for that category of table. If there are restrictions around PKs, perhaps that logic could be implemented in SimpleTable, such as only enabling this if the table has a single PK (or whatever the restriction needs to be...).
> 
> The problem is almost every table extends SimpleTable.
> 
> Here is the relevant PR and you will find more discussion around this. (Matt said in a comment" I think SimpleTable supports MERGE by default (this is sometimes a problem). We could leave it this way by default, and try to track down the cases where we actually want MERGE to work, but there might be a lot of usages.")
> 
> [LabKey/platform#3949](https://github.com/LabKey/platform/pull/3949)

@labkey-matthewb: Are there general patterns in which merge/replace does NOT work? If the table has a single PK, is there any reason it wouldnt normally work? I get that there are lots of permutations that occur in tables that inherit from SimpleTable, but it just seems like it would be useful if these new features could default to being available for a reasonable category of tables, without needing to opt-in one-by-one. Perhaps there should be a second base class that inherits from SimpleTable? Maybe there could be a constructor argument or configuration that is performed in SimpleUserSchema based on key fields?

bbimber:
@labkey-martyp Thanks for the EHRCustomPermissionsTable; however, I dont think that is the right place to set this. I'd still like to know this answer: if a basic module-provided table has one simple PK, is there any reason not to turn on MERGE/REPLACE/UPSERT? 

If there's no good reason not to, I think we should set those behaviors more comprehensively than EHRCustomPermissionsTable (whose purpose is really just permissions, like the name implies). Based on a quick look, it seems like we could: 1) implement something in SimpleUserSchema that checks the PK and conditionally turns them on (although maybe @labkey-matthewb / @XingY do not want this), 2) Add a new ExtendedSimpleTable class in LDK that sets those, and make CustomPermissionsTable, ContainerScopedTable, and SharedDataTable inherit from it, or 3) Set these values using DefaultTableCustomizer. 

If there are more complicated reasons to control when we do/dont enable this, and if the logic is too complicated to infer from the TableInfo itself, then the TableCustomizer layer is probably the best place to enable that, where we could opt-in tables by name.

XingY:
> @labkey-martyp Thanks for the EHRCustomPermissionsTable; however, I dont think that is the right place to set this. I'd still like to know this answer: if a basic module-provided table has one simple PK, is there any reason not to turn on MERGE/REPLACE/UPSERT?
> 
> If there's no good reason not to, I think we should set those behaviors more comprehensively than EHRCustomPermissionsTable (whose purpose is really just permissions, like the name implies). Based on a quick look, it seems like we could: 1) implement something in SimpleUserSchema that checks the PK and conditionally turns them on (although maybe @labkey-matthewb / @XingY do not want this), 2) Add a new ExtendedSimpleTable class in LDK that sets those, and make CustomPermissionsTable, ContainerScopedTable, and SharedDataTable inherit from it, or 3) Set these values using DefaultTableCustomizer.
> 
> If there are more complicated reasons to control when we do/dont enable this, and if the logic is too complicated to infer from the TableInfo itself, then the TableCustomizer layer is probably the best place to enable that, where we could opt-in tables by name.

It's possible that the PK for a table is an auto increment rowId, while during merge, some other combination of columns are provided as unique key. Auto integer is seldom provided during import.

The idea of using TableCustomizer make sense. It could work as an additional way to opt-in, on top of the existing TableInfo approach. It cannot replace the current TableInfo approach because not all tables can define a customizer, for example, a dataclass, sample type or list.

bbimber:
@XingY: point taken on auto-incrementing; however, arent these two questions separable?

1) Does the table theoretically support MERGE, in which case setting "setAllowedInsertOption(QueryUpdateService.InsertOption.MERGE);" is reasonable

2) Did the user submit a valid set of data to properly perform a MERGE?

Even if the table has an autoincrementing RowId, is there any reason QueryUpdateService.InsertOption.MERGE needs to be false? I might simply be missing something, but I do not currently see any reason why the vast majority of module-provided table should not have the following set as defaults, as @labkey-martyp is proposing here for one specific sub-class of these tables:
```
setAllowedInsertOption(QueryUpdateService.InsertOption.MERGE);
setAllowedInsertOption(QueryUpdateService.InsertOption.REPLACE);
setAllowedInsertOption(QueryUpdateService.InsertOption.UPSERT);
```

If this is not answered here, can one of you do what should be a <5 minute interactive phone/webmeeting? I think this would answer this much quicker than iterating more on this thread.

XingY (approved):

XingY:
> @XingY: point taken on auto-incrementing; however, arent these two questions separable?
> 
> 1. Does the table theoretically support MERGE, in which case setting "setAllowedInsertOption(QueryUpdateService.InsertOption.MERGE);" is reasonable
> 2. Did the user submit a valid set of data to properly perform a MERGE?
> 
> Even if the table has an autoincrementing RowId, is there any reason QueryUpdateService.InsertOption.MERGE needs to be false? I might simply be missing something, but I do not currently see any reason why the vast majority of module-provided table should not have the following set as defaults, as @labkey-martyp is proposing here for one specific sub-class of these tables:
> 
> ```
> setAllowedInsertOption(QueryUpdateService.InsertOption.MERGE);
> setAllowedInsertOption(QueryUpdateService.InsertOption.REPLACE);
> setAllowedInsertOption(QueryUpdateService.InsertOption.UPSERT);
> ```
> 
> If this is not answered here, can one of you do what should be a <5 minute interactive phone/webmeeting? I think this would answer this much quicker than iterating more on this thread.

The changes Marty made that's scoped to EHR tables makes sense. Other flavor of changes would work as well, as long as it's scoped to EHR. As for changes at SimpleUserSchema level, this is not something that we had planned to tackle at the moment. Maybe something for the future when we have cleaned up all instances that extends from it.

bbimber:
@XingY: thank you for you replies, but i'm going to ask this question elsewhere. Point taken on not implementing product-wide in SimpleUserSchema, but for the reasons I listed above I dont think adding this solely to EHRCustomPermissionsTable makes sense else.

labkey-martyp:
I synced with @XingY about this and just going with the minimal change for this use case for now.

PR #623, opened 2023-06-28
Hook for setting additional fields during bulk edit for custom fields

ankurjuneja:
#### Rationale
This PR provides a hook to support additional fields which should be set in the background during bulk edit in data entry forms.

#### Related Pull Requests
* https://github.com/LabKey/tnprc_ehr/pull/1225

### DIFF
--- a/ehr/resources/web/ehr/panel/BulkEditPanel.js
+++ b/ehr/resources/web/ehr/panel/BulkEditPanel.js
@@ -167,8 +167,17 @@ Ext4.define('EHR.panel.BulkEditPanel', {
         Ext4.Array.forEach(this.records, function(r){
             r.set(values);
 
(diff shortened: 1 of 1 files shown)


ehr/resources/web/ehr/panel/BulkEditPanel.js
```diff
@@ -167,6 +167,12 @@ Ext4.define('EHR.panel.BulkEditPanel', {
         Ext4.Array.forEach(this.records, function(r){
             r.set(values);
 
+            if (this.getBoundRecord() && this.getBoundRecord().setAdditionalFields) {
```
labkey-jeckels: Let's call this `bulkPropertyAdditionalFields` or something like that. Prefixing with `set` makes it sounds like it's a function. Also, please add a comment describing the use case and an example usage for those who don't have access to the

labkey-jeckels (approved):

PR #627, opened 2023-07-17
Issue 46960: EHR pedigree calculations using --vanilla flag which doesn't align with R package locations on hosted solutions

ankurjuneja:
#### Rationale
--vanilla flag used in EHR pedigree calculations is causing Rscript to look in .Library for packages instead of .Library.site which is used on hosted servers. According to the docs at https://colinfay.me/intro-to-r/appendix-b-invoking-r.html, it bundles together a bunch of other command-line flags like –no-save, –no-environ, –no-site-file, –no-init-file and –no-restore. Out of those, we can skip  –no-site-file as the --no-site-file flag skips the execution of site-specific initialization files, including the loading of site-wide environment variables and settings.

By avoiding the --no-site-file flag, Rscript will load the appropriate packages from the .Library.site directory based on the site-wide settings. This will allow us to utilize the site library instead of the user library for package loading.

#### Changes
* replace --vanilla flag with --no-save, --no-environ, --no-init-file and --no-restore

### DIFF
--- a/ehr/src/org/labkey/ehr/pipeline/GeneticCalculationsRTask.java
+++ b/ehr/src/org/labkey/ehr/pipeline/GeneticCalculationsRTask.java
@@ -126,7 +126,10 @@ public class GeneticCalculationsRTask extends WorkDirectoryTask<GeneticCalculati
 
         List<String> args = new ArrayList<>();
         args.add(exePath);
-        args.add("--vanilla");
+        args.add("--no-save"); // don't save workspace
+        args.add("--no-environ"); // Do not read any user file to set environment variables.
+        args.add("--no-init-file"); // Do not read the user’s profile at startup.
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):

labkey-martyp:
@labkey-jony you did a lot of work to install R libraries on our hosted solution around this vanilla flag. Do you foresee any issues removing it now?

bbimber:
FWIW, I dont believe there is any particularly good reason "--vanilla" was ever needed to be baked into this code. Allowing site admins more flexibility, especially since R is probably going to execute as a single-purpose service account, makes sense to me. I dont think this matters a ton one way or the other, but if anything I might omit these (but dont see a problem if they are kept):

args.add("--no-environ"); // Do not read any user file to set environment variables.
args.add("--no-init-file"); // Do not read the user’s profile at startup.

PR #638, opened 2023-08-11
Set phantom explicitly for new records

ankurjuneja:
#### Rationale
When copy previous request/task button is used in data entry forms containing tables and study datasets. The saveRows commands built for tables are for update as the new row is not marked as phantom for non-datasets. This PR explicitly marks the records as phantom.

### DIFF
--- a/ehr/resources/web/ehr/window/CopyTaskWindow.js
+++ b/ehr/resources/web/ehr/window/CopyTaskWindow.js
@@ -261,6 +261,8 @@ Ext4.define('EHR.window.CopyTaskWindow', {
 
                             var model = serverStore.addServerModel({});
                             model.set(obj);
+                            // this is a new record, so we need to mark it as phantom
+                            model.phantom = true;
 
                             this.toAddMap[serverStore.storeId].push(model);
                         }, this);


labkey-jeckels (approved):

bbimber:
@ankurjuneja: should this same change go into onprc_ehr/CopyTaskWindow, ~line 269?

ankurjuneja:
@bbimber if the data entry form using onprc_ehr/CopyTaskWindow has non-datasets grid/form sections, then yes.

bbimber:
@ankurjuneja: 

1) Datasets are the dominant use-case, but queries like MiscCharges do exist in various form types and those are not a dataset. 

2) I know this code pre-dates this PR, but the fact this code needs to set or even be aware of 'phantom' feels like something is bleeding out internally within the store in a way it should not. 

When I reviewed this I noticed a unrelated issue in this code and I'm proposing a fix here: https://github.com/LabKey/ehrModules/pull/639

That addresses the double-add issue (described on that PR), and I propose an (admittedly not that much better), but marginally more declarative way to address phantom and keep it more internal within store itself. That would be an alternative to this PR. 

I also wonder why this code isnt adding the new records to the client store (which is more in line with how grids and forms work), as opposed to the serverStore, but I think that's out of scope to dig into. .

PR #639, opened 2023-08-12
Do not double-add records to DataEntryStore in CopyTaskWindow

bbimber:
This is related to: https://github.com/LabKey/ehrModules/pull/638

When reviewing that PR, I noticed that CopyTaskWindow was double-adding new records to the serverStore. Previously, CopyTaskPanel would query the server, and in the successCallback for each query, it would call serverStore.addServerModel(). This ends up creating that record in the store; however, it also builds an internal map of those records (this.toAddMap). After the MultiRequest returns, onLoad is called and then the code re-adds all these records to the server store. I can only assume that some internal checks within the store prevent double-adding, or maybe there was a race condition happening around how and when StoreCollection.transformServerToClient is called/triggered, but this pattern doesnt make sense.

This PR:
- Separates DataEntryServerStore.addClientModel and createClientModel methods.
- Uses createClientModel here, avoiding adding records twice.
- Related to #638, this makes phantom more declarative as an argument in createClientModel, which it is marginally more verbose way to flag upstream code to be aware of phantom, and keeps 'phantom' internal within the store.
- Related to https://github.com/LabKey/onprcEHRModules/pull/759, some time ago ONPRC simply copy/pasted this entire file to make a minor change in behavior. This PR adds a 'defaultQCStateLabel' property to the window, which lets ONPRC set this, and avoids them copying the entire file.

### DIFF
--- a/ehr/resources/web/ehr/data/DataEntryServerStore.js
+++ b/ehr/resources/web/ehr/data/DataEntryServerStore.js
@@ -618,16 +618,28 @@ Ext4.define('EHR.data.DataEntryServerStore', {
         //}
     },
 
-    //creates and adds a model to the provided server store, handling any dependencies within other stores in the collection
-    addServerModel: function(data){
+    // creates a model for the provided server store, handling any dependencies within other stores in the collection
+    createServerModel: function(data, doesNotExistOnServer){
         data = data || {};
         if (EHR.debug)
             console.log('creating server model');
         var model = this.createModel(data);
+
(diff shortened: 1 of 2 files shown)


ankurjuneja (approved):

bbimber:
@ankurjuneja: so is this good to go? does anybody else need to weigh in on this?

ankurjuneja:
@bbimber testing it locally works, so I am good. Thanks for checking.

bbimber:
OK, thanks.

PR #644, opened 2023-08-23
SimpleQueryForm double title

labkey-martyp:
#### Rationale
ExtJS form is expecting null value for title in order to not display panel header. The page already has a title on it.

#### Changes
* Preserve nulls

### DIFF
--- a/ehr/api-src/org/labkey/api/ehr/dataentry/SingleQueryFormSection.java
+++ b/ehr/api-src/org/labkey/api/ehr/dataentry/SingleQueryFormSection.java
@@ -16,6 +16,7 @@
 package org.labkey.api.ehr.dataentry;
 
 import org.json.JSONObject;
+import org.labkey.api.util.JsonUtil;
 import org.labkey.api.view.template.ClientDependency;
 
 import java.util.HashMap;
@@ -43,10 +44,10 @@ public class SingleQueryFormSection extends SimpleFormSection
         Map<String, Object> bindConfig = new HashMap<>();
         bindConfig.put("createRecordOnLoad", true);
         formConfig.put("bindConfig", bindConfig);
-        formConfig.put("title", JSONObject.NULL);
-        formConfig.put("label", JSONObject.NULL);
+        formConfig.put("title", null);
+        formConfig.put("label", null);
         formConfig.put("maxItemsPerCol", 8);
-        ret.put("formConfig", formConfig);
+        ret.put("formConfig", JsonUtil.toJsonPreserveNulls(formConfig));
 
         return ret;
     }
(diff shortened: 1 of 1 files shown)


bbimber:
hey @labkey-martyp, I was a little worried the JSON conversion was going to result in obscure, hard to find issues. If I understand your comment correctly, does this mean something on the client is specifically looking for NULL and not undefined? What code is the problem there?

It's too bad JSONObject doesnt handle the null -> JSONObject.NULL conversion for you when adding a Map to a JSONObject.

labkey-martyp:
> hey @labkey-martyp, I was a little worried the JSON conversion was going to result in obscure, hard to find issues. If I understand your comment correctly, does this mean something on the client is specifically looking for NULL and not undefined? What code is the problem there?
> 
> It's too bad JSONObject doesnt handle the null -> JSONObject.NULL conversion for you when adding a Map to a JSONObject.

@bbimber I tracked it back as far as the Ext.form.Panel, so it appears to be something in the ExtJS library that will not show the panel header if title is explicitly set to null. It doesn't specifically call it out in documentation.

Eh, actually we do have something as you're suggesting in JsonUtils. That might be a better refactor.

bbimber:
@labkey-martyp, FWIW, it's this Ext4.apply() that is causing this, not something in Ext4 itself:

https://github.com/LabKey/ehrModules/blob/f78f28abc639953ad7894831d620f19bcec63b36/ehr/resources/web/ehr/panel/DataEntryPanel.js#L364

The title defaults to section.label (i.e., the table's name). If the title property is present in section.formConfig, it will overwrite 'title' with NULL in the apply() call. If section.formConfig lacks the title property, this doesnt happen and section.label persists as the form's title. 

There is a lot of object merging within EHR's client config, and this pattern (expecting a NULL to unset something), probably isnt all that rare.

labkey-martyp:
@bbimber got it thanks. I just started at SimpleDataEntryFormPanel and started looking there, I didn't realize the parent form was manipulating the title.

Yeah there are several places where we're converting maps to JSON. We could do a bigger refactor to find all those and preserve nulls but that seems a little heavy handed adding a bunch of null values where they're probably not needed. This JSON change has been out in production for a while now, I think we've covered a lot of ground finding null issues.

I agree it would be nice if JSONObject itself had a method to add a map and preserve nulls.

bbimber:
@labkey-martyp, yeah I do not recommend globally trying to replace all instance when code is setting NULL in a JSONObject or re-instating a second subclass or something that drastic. I would bet there are many more cases like this though.

labkey-jeckels (approved):

labkey-adam:
@bbimber @labkey-martyp I'm late to the party, but FYI: we do have `JsonUtil.toJsonPreserveNulls(Map<String, Object> map)`

bbimber:
Thanks, I think Marty is using that. That helper certainly is useful case-by-case, but ensuring there are not uses like this one is a little tricky. 

FWIW, I tried searching for this regex: 

put\("(.)*, null\);

which finds about a dozen instances of adding a null to a map; however, I didnt see any instances that became JSONObjects or needed updating like this. That regex probably not the only possible pattern, but I bet would get a lot of these.

PR #649, opened 2023-08-30
Improve stability of VL Test

bbimber:
An example failure is here: https://teamcity.labkey.org/buildConfiguration/LabKey_237Release_External_Discvr_ExternalModulesTestSqlserver/2611285?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildTestsSection=true&expandBuildChangesSection=true. I'm pretty sure this will improve that. The same field is set on line 444 using setByDisplayValue(), which should be a more stable way to do this.


bbimber:
@labkey-bpatel: while there are test failures, the ViralLoadAssayTest is passing.

labkey-jeckels (approved):

PR #663, opened 2023-09-25
New non-client EHR module for react development

ankurjuneja:
#### Rationale
This is a non-client EHR module to use for automated testing and for demoing the setup and use of the new React pages.

#### Changes
* new basic EHR App module


EHR_App/test/src/org/labkey/test/tests/EHR_AppTest.java
```diff
@@ -0,0 +1,124 @@
...
+    @Override
+    protected String getAnimalHistoryPath()
+    {
+        return null;
+    }
+
+    @Before
+    public void preTest()
+    {
+        goToProjectHome();
```
labkey-martyp: I know we do this in other places, but shouldn't this be goToEHRFolder?

EHR_App/test/src/org/labkey/test/tests/EHR_AppTest.java
```diff
@@ -0,0 +1,124 @@
...
+    private static final String PROJECT_NAME = "EHR App";
+    private static final String FOLDER_NAME = "EHR";
+    private final EHRAppTestSetupHelper _setupHelper = new EHRAppTestSetupHelper(this, getProjectName(), FOLDER_NAME, getModulePath(), getContainerPath());
+
+    @Override
+    protected String getProjectName()
+    {
+        return PROJECT_NAME;
+    }
+
```
labkey-martyp, line 30:
```suggestion
    @Override
    protected boolean doSetUserPasswords()
    {
        return true;
    }
```

The built-in EHR permissions tests should pass if you add this function

labkey-martyp (changes requested):
Looks good. A couple requests.
- Can you add QCState to the sample data TSVs with a value of Completed. That makes it all public data and things just work better.
- Can we get all EHR_AppTests passing (mostly built-in EHR tests). Eventually we'll be adding this to the EHR suite. Even if we have to override them and add a comment why they are overridden.
- It would be great to add some data entry forms that can insert data with Completed QCState. Is that something we can get into this PR or would it be better to make that a follow on story?

labkey-martyp (approved):
Looks great. Thanks Ankur!

PR #666, opened 2023-09-26
Cases update and override

labkey-martyp:
#### Rationale
Couple updates to facilitate cases implementation. 
- AnimalField fix to account for auto-upper casing of Ids in forms - ensure the animal lookup is using getValue instead of accessing value directly (gets the upper cased values in this case). 
- Make code field in sedation window overridable to facilitate different snomed codes

#### Related Pull Requests
* https://github.com/LabKey/johnsHopkinsEHRModules/pull/112

#### Changes
* Use getValue instead of value on AnimalField blur
* Move code field into separate function in sedation window

### DIFF
--- a/ehr/resources/web/ehr/form/field/AnimalField.js
+++ b/ehr/resources/web/ehr/form/field/AnimalField.js
@@ -21,7 +21,7 @@ Ext4.define('EHR.form.field.AnimalField.js', {
 
         // Verify one more time when the field loses focus as change handler can be a little flaky
         this.on('blur', function(field){
-            this.fireEvent('animalchange', field.lastValue || field.value);
+            this.fireEvent('animalchange', field.getValue());
         }, this);
     },
 
--- a/ehr/resources/web/ehr/form/field/AnimalField.js
+++ b/ehr/resources/web/ehr/form/field/AnimalField.js
@@ -21,7 +21,7 @@ Ext4.define('EHR.form.field.AnimalField.js', {
 
         // Verify one more time when the field loses focus as change handler can be a little flaky
         this.on('blur', function(field){
-            this.fireEvent('animalchange', field.value);
+            this.fireEvent('animalchange', field.lastValue || field.value);
         }, this);
     },
 
--- a/ehr/resources/web/ehr/window/SedationWindow.js
+++ b/ehr/resources/web/ehr/window/SedationWindow.js
@@ -126,6 +126,47 @@ Ext4.define('EHR.window.SedationWindow', {
         }]
     },
 
+    getCodeField: function(key){
+        return {
+            xtype: 'combo',
+            key: key,
+            fieldName: 'code',
+            valueField: 'code',
+            displayField: 'displayField',
+            value: 'E-70590',
+            store: {
+                type: 'store',
+                fields: ['code', 'displayField', 'dosage'],
+                data: [
+                    {code: 'E-70590', displayField: 'Ketamine', dosage: 10},
+                    {code: 'E-YY992', displayField: 'Telazol', dosage: 3},
+                    {code: 'none', displayField: 'None', dosage: null}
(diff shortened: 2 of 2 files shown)


ehr/resources/web/ehr/form/field/AnimalField.js
```diff
@@ -21,7 +21,7 @@ Ext4.define('EHR.form.field.AnimalField.js', {
 
         // Verify one more time when the field loses focus as change handler can be a little flaky
         this.on('blur', function(field){
-            this.fireEvent('animalchange', field.value);
+            this.fireEvent('animalchange', field.lastValue || field.value);
```
bbimber: @labkey-martyp: what's this about, and what code sets lastValue?

labkey-martyp: It's set in Field.js in ext code. It's the most recent value in the field. For an auto-upper casing animal id field, it was not catching the upper case value.

labkey-martyp: https://github.com/LabKey/platform/blob/325233000a2cdda5c3c238357211c28e698ec632/api/webapp/ext-4.2.1/src/form/field/Field.js#L317

ankurjuneja (approved):

ehr/resources/web/ehr/form/field/AnimalField.js
bbimber: Thanks for the pointer. I get how it works, but it's still relying on a non-public Ext value. Granted, I dont think anyone is upgrading Ext versions, but still not ideal.

If I understand you, the reason for adding the blur is that change alone wasnt sufficient? Do you have a good sense as far as when change isnt reliable?

Granted, it's a different sort of workaround, but what if you simply added {delay: 100} to your blur handler? Perhaps that gives enough time for the listener that does the uppercase to run? What code is executing the upper case? Perhaps whatever executes that could hook into a different point more upstream?

The reason I am generally leery with mucking into things like this is unintended consequences. Are there unanticipated scenarios where field.lastValue in out of date? Should we actually always defer to that instead of value? 

Based on the code you linked, if lastValue is getting set, then the change event is immediately fired anyway, isnt it? It does seem very possible for it to get stale (at least in the narrow window around a change event).

Whatever the solution is here, this seems like a situation where verifying behavior could be useful. Is there something amenable to get tested using an LDK.Assert as far as what you expect from lastValue/value? In my experience, there are enough possible permutations in usage that it might be useful to log to the server whenever those expectations are not met. It's the only effective method I've found to work out some of these context/timing related things.

You alluded to upper case being a problem. If AnimalIdUpperField.js is where this is used, could that class do something more direct here? Perhaps override field.getRawValue() or something very low level that ensures string values are always upper-cased?

Just throwing out some ideas here.

labkey-martyp: Thanks for the brainstorming @bbimber. Looking at this again, I can just call getValue instead of accessing value directly. That function is what is overridden for AnimalIdUpperField

labkey-martyp: And to answer your question about the blur event. Yeah there are just some cases depending on how fast you type and the response time I guess, where the animal Id validation was for a partial Id, even though the full Id was typed. This is especially noticeable in tests, but would also occasionally repro when manually entering the Id.

bbimber: My comment on these very context dependent race or race-like conditions remains: the whole reason all the LDK.Assert stuff exists is because gathering real-world usage data about how your app is behaving really does matter. The overhead is minimal and it provides a simple way to gather information on whether the assumptions you are making about some behavior really work or not.

In this instance, glad AnimalIdUpperField.getValue works

PR #667, opened 2023-09-30
Issue 47265 - Data validation queries for dams and sires

ankurjuneja:
#### Rationale
See related

#### Related Pull Requests
* https://github.com/LabKey/ehrModules/pull/534

#### Changes
* hidden html report
* sql quereis to identify outliers

PR #670, opened 2023-10-02
fix ehr data validation queries

ankurjuneja:
#### Rationale
This PR fixes the related PR for the centers that store dam and sire in different dataset than demographics

#### Related Pull Requests
* https://github.com/LabKey/ehrModules/pull/667

#### Changes
* use Id.parents.dam in place of dam from demographics

### DIFF
--- a/ehr/resources/queries/study/animalIdsAsDamAndSire.sql
+++ b/ehr/resources/queries/study/animalIdsAsDamAndSire.sql
@@ -1,14 +1,14 @@
 
-SELECT DISTINCT dam AS Id,
+SELECT DISTINCT Id.parents.dam AS Id,
                 gender,
                 species
 FROM demographics
-WHERE dam IN (SELECT sire FROM demographics)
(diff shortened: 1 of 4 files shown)


labkey-jeckels (commented):
It might speed up the queries to use `study.demographicsParents` as the base query instead of starting with `study.demographics` and doing two separate lookup-based joins to it via `Id.parents`

labkey-martyp:
> It might speed up the queries to use `study.demographicsParents` as the base query instead of starting with `study.demographics` and doing two separate lookup-based joins to it via `Id.parents`

@ankurjuneja How's the perf after your refactors Ankur. Want to investigate Josh's suggestion here?

ankurjuneja:
@labkey-martyp After fixing the slowest one, all 4 queries are loading in ~40-50 seconds

labkey-martyp (approved):
Looks good. If you're seeing one or two of these queries taking the majority of the load time, probably worth looking into Josh's suggestion on that one.

PR #673, opened 2023-10-06
Issue 48803: Update ehr schema on SQLServer to use NVARCHAR and upgrade postgres from varchar(4000) to text

ankurjuneja:
#### Rationale
Based on the long-standing recommendation https://www.labkey.org/Documentation/wiki-page.view?name=sqlScriptConventions and recent mangling characters in requests table's columns in TNPRC, this PR updates the ehr schema on SQLServer to use NVARCHAR, which supports non-ASCII characters and transforms it's postgres equivalent's from varchar(4000) to text.

#### Changes
* sql server upgrade script
* postgres upgrade script


ankurjuneja:
@labkey-jeckels please hold off on code review, need to delete auto generated constraints through some other mechanism

bbimber:
@ankurjuneja: would this make sense to implement as java update code, where you could have some java code iterate every single table in the schema, inspect each column, inspect indexes, and execute the appropriate sql?

labkey-jeckels:
@bbimber Java upgrades are great for cases when we truly need them, but they're also more fragile. This case is relatively low risk, but I still prefer to avoid them when possible.

That said, in the spirit of your suggestion, using code to generate a static upgrade script which can then be checked in and run reliably in the regular upgrade sequence could be a useful technique.

bbimber:
hi @labkey-jeckels, yeah, i think we both get that. especially if this is going to be context dependent, like different servers having different indexes, then it seems like where it might be useful...

ankurjuneja:
@labkey-jeckels this is ready for review

ehr/resources/schemas/dbscripts/sqlserver/ehr-23.000-23.001.sql
```diff
@@ -0,0 +1,432 @@
...
+
+ALTER TABLE ehr.protocolexemptions ALTER COLUMN protocol NVARCHAR(100);
+ALTER TABLE ehr.protocolexemptions ALTER COLUMN exemption NVARCHAR(200);
+ALTER TABLE ehr.protocolexemptions ALTER COLUMN remark NVARCHAR(MAX);
+GO
+
+ALTER TABLE ehr.protocolprocedures ALTER COLUMN protocol NVARCHAR(200) NOT NULL;
+ALTER TABLE ehr.protocolprocedures ALTER COLUMN procedurename NVARCHAR(200);
+ALTER TABLE ehr.protocolprocedures ALTER COLUMN code NVARCHAR(100);
+ALTER TABLE ehr.protocolprocedures ALTER COLUMN frequency NVARCHAR(2000);
```
labkey-jeckels: This was originally a VARCHAR(200)
```suggestion
ALTER TABLE ehr.protocolprocedures ALTER COLUMN frequency NVARCHAR(200);
```

labkey-jeckels (approved):

PR #685, opened 2023-11-11
Hook to override default lab work type

ankurjuneja:
#### Rationale
This PR provides a way to register lab work override in center specific modules

#### Related Pull Requests
* https://github.com/LabKey/johnsHopkinsEHRModules/pull/129


labkey-martyp (approved):

ehr/src/org/labkey/ehr/EHRServiceImpl.java
```diff
@@ -1050,4 +1051,14 @@ public void appendSNOMEDCols(AbstractTableInfo ti, String displayColumnName, Str
             ti.addColumn(simpleRawCol);
         }
     }
+
+    public void registerLabWorkOverrides(Module module, String fromType, LabworkType toType)
+    {
+        _labWorkOverrides.put(fromType, Pair.of(module, toType));
```
labkey-jeckels, line 1057: While we don't have conflicting overrides yet, this doesn't feel right in terms of how it treats the Module argument. Whichever module registers last will win. And the usage in LabworkManager totally ignores the Module.

I think _labworkOverrides should be a Map<String, Map<Module, LabworkType>>. And then when we look at the overrides, we should only apply those for modules that are active in the container.

labkey-jeckels (commented):
I didn't review in time before this was merged, but please take a look at my feedback for a followup PR

PR #687, opened 2023-11-19
check for active modules when returning registered labwork types overrides

ankurjuneja:
#### Rationale
This PR addresses the code review comment https://github.com/LabKey/ehrModules/pull/685#discussion_r1398006704 in related PR.

#### Related Pull Requests
* https://github.com/LabKey/ehrModules/pull/685

### DIFF
--- a/ehr/src/org/labkey/ehr/history/LabworkManager.java
+++ b/ehr/src/org/labkey/ehr/history/LabworkManager.java
@@ -59,17 +59,20 @@ public class LabworkManager
     public Collection<LabworkType> getTypes(Container c)
     {
         List<LabworkType> result = new ArrayList<>(_types.size());
-        Map<String, Pair<Module, LabworkType>> labworkTypeOverrides = EHRServiceImpl.get().getLabWorkOverrides();
+        Map<String, Map<Module, LabworkType>> labworkTypeOverrides = EHRServiceImpl.get().getLabWorkOverrides();
         Set<Module> activeModules = c.getActiveModules();
         for (LabworkType type : _types)
         {
             if (type.isEnabled(c))
             {
-                if (labworkTypeOverrides.containsKey(type.getName()) &&
-                    activeModules.contains(labworkTypeOverrides.get(type.getName()).first))
+                if (labworkTypeOverrides.containsKey(type.getName()))
                 {
-                    Pair<Module, LabworkType> override = labworkTypeOverrides.get(type.getName());
-                    result.add(override.getValue());
+                    labworkTypeOverrides.get(type.getName()).forEach((module, labworkType) -> {
+                        if (activeModules.contains(module))
+                        {
+                            result.add(labworkType);
+                        }
+                    });
                 }
                 else
                 {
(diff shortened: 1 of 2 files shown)


ehr/src/org/labkey/ehr/history/LabworkManager.java
```diff
@@ -59,11 +60,13 @@ public Collection<LabworkType> getTypes(Container c)
     {
         List<LabworkType> result = new ArrayList<>(_types.size());
         Map<String, Pair<Module, LabworkType>> labworkTypeOverrides = EHRServiceImpl.get().getLabWorkOverrides();
+        Set<Module> activeModules = c.getActiveModules();
         for (LabworkType type : _types)
         {
             if (type.isEnabled(c))
             {
-                if (labworkTypeOverrides.containsKey(type.getName()))
+                if (labworkTypeOverrides.containsKey(type.getName()) &&
```
bbimber: If LabworkType already has an isEnabled(Container) method, why not just use that? It's more object-oriented and future proof if there are other reasons to conditionalize when an override should happen 

Something like:
```
if (labworkTypeOverrides.containsKey(type.getName()) && labworkTypeOverrides.get(type.getName()).second.isEnabled(c))
```
It would probably be useful to have a base LabworkType (or LabworkTypeOverride) class that accepts a module in the constuctor and implements isEnabled with this kind of check against active modules. This would mean implementations in center modules wouldnt need to worry about this.

bbimber (changes requested):

labkey-jeckels (changes requested):
This doesn't look like it addresses my comment about changing _labworkOverrides to a Map<String, Map<Module, LabworkType>>

ehr/src/org/labkey/ehr/history/LabworkManager.java
ankurjuneja: DefaultLabworkType already accepts module and isEnabled checks if this is enabled for current module. ehrModules has default implementations for Labworktypes that extend DefaultLabworkType  and ehrModules is enabled in center specific modules too. Implementing isEnabled in center specific modules would just add to the list and won't replace just the targeted LabworkType which is the intention.

bbimber: Hi @ankurjuneja: Yes, I wrote the second paragraph without having the code in front of me. Please disregard that part of it.

Nonetheless, as you point out, LabworkTypes are instantiated with the owning module. I have this code in front of me, and I actually have two more serious questions:

Does the concept of LabworkOverrides actually even need to exist? DataEntryForms and History and other similar EHR resources have exactly the same needs. You have core EHR supplying default implementations, but you want center-specific modules to be able to override them. 

Take a look at how DataEntryManager deals with this here: https://github.com/LabKey/ehrModules/blob/8da82eef5e683daeecb4b452557dc7328952539f/ehr/src/org/labkey/ehr/dataentry/DataEntryManager.java#L85. Because modules are initialized in dependency order, their the resources registered using the existing standard registerLabworkType() method will be in order. 

Can't you just apply the same kind of logic to labworkTypes, and therefore get rid of the entire concept of a separate method for labworkTypeOverrides? In order to override a default LabworkType, all the center-specific module needs to do is create one with the same name, and it would override the core one automatically, without needing a redundant entry point and parallel scheme:

```
    public Collection<LabworkType> getTypes(Container c)
    {
        Map<String, LabworkType> map = new HashMap<>();
        for (LabworkType type : _types)
        {
            if (type.isEnabled(c))
                map.put(type.getName(), type);
        }

        return Collections.unmodifiableCollection(map.values());
    }

```

PR #713, opened 2024-01-24
Collapsible Data Entry

labkey-martyp:
#### Rationale
EHR data entry forms can be very noisy, out of sequence with workflows and unclear for new users where to start entering data. This PR provides an optional way to collapse sections of the data entry forms to minimize the visible options when first navigating to the form. Headers of collapsed sections are visible with lower opacity and can be clicked to expand. Any sections with data entered will maintain full opacity of the header even when collapsed to ensure users do not forget they entered data in that section. Loading existing tasks with data will automatically expand any sections with data. 

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/1808

#### Changes
* Add CollapsibleDataEntryPanel plugin
* Update form and grid panels to add plugin if "collapsible" field is defined/true in the formConfig
* Opacity css update
* Test update


labkey-jeckels (approved):
Sounds like a nice bit of UI

ankurjuneja (approved):
I've tested this enhancement locally, and I must say it's rad.

PR #717, opened 2024-02-01
Add event to expand collapsible panels

labkey-martyp:
#### Rationale
Need event to fire when panel should be expanded

#### Related Pull Requests
* https://github.com/LabKey/johnsHopkinsEHRModules/pull/187

#### Changes
* Add expand panel event

### DIFF
--- a/ehr/resources/web/ehr/plugin/CollapsibleDataEntryPanel.js
+++ b/ehr/resources/web/ehr/plugin/CollapsibleDataEntryPanel.js
@@ -8,6 +8,11 @@ Ext4.define('EHR.plugin.CollapsibleDataEntryPanel', {
 
     init: function(panel){
 
+        // Don't apply this to sub-panels in other windows/panels
+        if (!panel.border) {
+            return;
+        }
+
         panel.collapsible = true;
 
         if ((!panel.store || panel.store.getCount() === 0) && panel.formConfig?.initCollapsed) {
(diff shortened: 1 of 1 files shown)


ehr/resources/web/ehr/plugin/CollapsibleDataEntryPanel.js
```diff
@@ -14,6 +14,12 @@ Ext4.define('EHR.plugin.CollapsibleDataEntryPanel', {
             panel.collapsed = true;
         }
 
+        panel.onPanelDataChange = function() {
+            panel.expand();
+        }
+
+        panel.mon(panel, 'panelDataChange', panel.onPanelDataChange, {buffer: 500});
```
labkey-jeckels: I think the last argument will be interpreted at the scope and not the options. Can it be removed, or can we add a scope?

https://cdn.sencha.com/ext/gpl/4.2.1/docs/#!/api/Ext.util.Observable

`mon( item, ename, [fn], [scope], [options] ) `

labkey-jeckels (approved):

ehr/resources/web/ehr/plugin/CollapsibleDataEntryPanel.js
labkey-martyp: fixed thanks

PR #725, opened 2024-02-16
Sample and Animal Import Fixes

labkey-martyp:
#### Rationale
A few updates and fixes to support sample and animal imports.

#### Changes
- Update ETL to use the sourceContainerPath instead of dot notation schema
- Make the Animal Record column not required
- Don't change the name of Id alias column, just give it a label


EHR_SM/resources/etls/animal_template.xml
```diff
@@ -8,7 +8,7 @@
             <description>Copy Animal Data to Animal Source</description>
             <!-- For source, update folderPath to the folder path of your EHR folder using dot notation. queryName can be
```
labkey-jeckels, line 9: The comment here and a few lines lower is slightly out of date with the new placeholder value

labkey-jeckels (approved):

PR #726, opened 2024-02-18
Path-first URLs

labkey-adam:
#### Rationale
Path-first URLs are the "new" standard

### DIFF
--- a/EHR_App/resources/views/begin.html
+++ b/EHR_App/resources/views/begin.html
@@ -11,7 +11,7 @@ function createNavMenu(){
 
     Ext4.get(webpart.wrapperDivId).update(
         '<div style="max-width:920px">' +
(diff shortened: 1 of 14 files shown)


bbimber:
hey @labkey-adam: so does this imply the newer URL pattern of <containerPath>/<controller>-<action> is now favored over the old-style /<controller>/<containerPath>/<action>? Is the other style getting deprecated? There is probably a fair amount of the original style URLs across EHR and my lab's modules, since a lot of that was written before the new style existed.

labkey-adam:
> hey @labkey-adam: so does this imply the newer URL pattern of /- is now favored over the old-style ///? Is the other style getting deprecated? There is probably a fair amount of the original style URLs across EHR and my lab's modules, since a lot of that was written before the new style existed.

Newer style is preferred, but old style will be supported indefinitely. Since I was touching all the EHR repos to add script nonces I took the opportunity to search and replace the URLs pretty much everywhere as well.

labkey-matthewb (approved):

PR #730, opened 2024-02-28
Refactor kinship to allow other code to initiate import directly

bbimber:
This PR contains two main categories of changes. The motivation of these changes was to separate the process of computing the data for EHR genetics (e.g. kinship and inbreeding) from actually importing the data into the EHR. In total, the idea is that PRIMe-seq (which already has a mirrored copy of pedigree data), will execute the default EHR GeneticsCalculation pipeline. It will farm the computation to our cluster, which that server is already configured to do. When complete, this pipeline already saves the results as TSV files. I wrote a separate ETL that will be defined in ONPRC modules, which copies the resulting TSVs to a location visible to PRIMe, and then pings PRIMe via a new server-side action to cause PRIMe to take those TSVs and call EHRService.standaloneProcessKinshipAndInbreeding to actually import them.

The changes within EHR itself are primarily to refactor the portions of the code that import the TSVs be to static, allow it to be called separately, and expose this through EHRService. These changes are basically a refactor without touching much within the code itself.

When I got into the weeds of the R code, I noticed a number of other things that seemed worth cleaning up. These are not directly related to the importing of data, but should be broadly useful:

- I did some general style improvements in the R code, and removed some ancient strange patterns. This resulted in more code being touched than strictly needed, but most changes are minor and make the script a little more standard.

- I think I understand the intent behind "options(error = dump.frames)", but I dont think it was giving the result it should. In my hands, not only was this not writing anything to a file, the script did not die on errors. I'd argue it's a lot more useful if it dies when there's an error (such as not having kinship2 installed), rather than continue onward after errors and accumulate strange downstream errors. I tested this pretty thoroughly locally and I think R's default error handling (i.e. dont specify anything with options()) does as good a job as anything here. Here is an example (which requires a login), showing how the existing 23.7 code ignores R errors: https://prime-seq.ohsu.edu/pipeline-status/Internal/PMR/details.view?rowId=433466.

This was originally opened for 23.7 under #650. This PR backs out some of the changes originally part of that PR. Notably, this is not attempting to address the WNPRC issue around hybrid animals, nor does it try to perform sanity checking of minimum expected kinship coefficients in R.

### DIFF
--- a/ehr/resources/pipeline/kinship/populateKinship.r
+++ b/ehr/resources/pipeline/kinship/populateKinship.r
@@ -7,30 +7,24 @@
 # This R script will calculate and store kinship coefficients (aka. relatedness) for all animals in the colony.  This is a large, sparse matrix.
 # The matrix is converted into a very long 3-column dataframe (animal1, animal2, coefficient).  This dataframe is output to a TSV file,
 # which is normally imported into ehr.kinship by java code in GeneticCalculationsImportTask
-
-
-#options(echo=TRUE);
-options(error = dump.frames);
-library(methods);
-library(kinship2);
-library(getopt);
-library(Matrix);
-library(dplyr);
+options(error = dump.frames)
+library(kinship2)
+library(getopt)
+library(Matrix)
+library(dplyr)
 
 spec <- matrix(c(
-#'containerPath', '-c', 1, "character",
-#'baseUrl', '-b', 1, "character"
-'inputFile', '-f', 1, "character"
-), ncol=4, byrow=TRUE);
-opts = getopt(spec, commandArgs(trailingOnly = TRUE));
+    'inputFile', '-f', 1, "character"
+), ncol=4, byrow=TRUE)
+opts <- getopt(spec, commandArgs(trailingOnly = TRUE))
 
-allPed <- read.table(opts$inputFile, quote="\"");
-colnames(allPed)<-c('Id', 'Dam', 'Sire', 'Gender', 'Species');
+allPed <- read.table(opts$inputFile, quote="\"")
+colnames(allPed)<-c('Id', 'Dam', 'Sire', 'Gender', 'Species')
 
-is.na(allPed$Id)<-which(allPed$Id=="")
-is.na(allPed$Dam)<-which(allPed$Dam=="")
-is.na(allPed$Sire)<-which(allPed$Sire=="")
-is.na(allPed$Gender)<-which(allPed$Gender=="")
+allPed$Id[allPed$Id == ""] <- NA
+allPed$Id[allPed$Dam == ""] <- NA
+allPed$Id[allPed$Sire == ""] <- NA
+allPed$Id[allPed$Gender == ""] <- NA
 
 allPed$Species <- as.character(allPed$Species)
 allPed$Species[is.na(allPed$Species)] <- c('Unknown')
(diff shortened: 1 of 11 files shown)


bbimber:
@labkey-martyp: this is a simplification of the earlier PR. There's a JHU postgres failure, but it does not seem related to the genetics calculations.

labkey-martyp:
Thanks @bbimber let me test this out on data from the centers, otherwise it looks good.

bbimber:
Thanks @labkey-martyp. Like I posted on the support thread, please keep in mind that you can just run populateKinship standalone, where you replace the TSV loading with RLabkey/SelectRows. That should make it quick to point to any accessible center's server (even a production server) to non-invasive testing on real data.

ehr/src/org/labkey/ehr/pipeline/GeneticCalculationsImportTask.java
```diff
@@ -145,54 +147,62 @@ public RecordedActionSet run() throws PipelineJobException
...
-            processInbreeding();
-            processKinship();
+            PipelineJob job = getJob();
+            FileAnalysisJobSupport support = (FileAnalysisJobSupport) job;
 
-            if (GeneticCalculationsJob.isKinshipValidation())
+            processInbreeding(job.getContainer(), job.getUser(), support.getAnalysisDirectoryPath().toFile(), job.getLogger());
+            processKinship(job.getContainer(), job.getUser(), support.getAnalysisDirectoryPath().toFile(), job.getLogger());
+
+            if (Boolean.parseBoolean(getJob().getParameters().get("kinshipValidation")))
```
labkey-martyp: We're not adding this parameter in the job so this is never true. This can either go back to calling isKinshipValidation or add that value to the job parameters when it is created. Looks like that is in GeneticsCalculationsRunnable.

ehr/resources/pipeline/kinship/populateKinship.r
```diff
@@ -7,67 +7,70 @@
...
-    allRecordsForSpecies <- allPed[allPed$Species == species,]
+    allRecordsForSpecies <- allPed[allPed$Species %in% species,]
+    print(paste0('Processing species: ', species, ', with ', nrow(allRecordsForSpecies), ' IDs'))
+    if (nrow(allRecordsForSpecies) == 1) {
+        print('single record, skipping')
+        next
+    }
+
+    pctMissingSex <- sum(allRecordsForSpecies$Gender > 2) / nrow(allRecordsForSpecies)
+    if (pctMissingSex > 0.25) {
```
labkey-martyp: This check is already being done in fixParents, and it's only throwing a warning there and still performing the calculations. I don't think we want to stray from that. I think we can remove this check.

labkey-martyp (changes requested):
Ok I tested this out across centers. Looks good with just a couple requests. Check them out, let me know if you have any follow up on them.

ehr/resources/pipeline/kinship/populateKinship.r
bbimber: I thought I remembered fixParents throwing, but if you tried this and it's a warning then you're probably fine.

ehr/src/org/labkey/ehr/pipeline/GeneticCalculationsImportTask.java
bbimber: Yes, good point. I didnt back this out when I removed the other changes

bbimber:
Thanks @labkey-martyp. The last commit addresses those 2 comments.

labkey-martyp (approved):
Looks good. Thanks Ben.

PR #738, opened 2024-03-15
Resolve LabworkTypes in module dependency order to remove need for registerLabworkOverrides

bbimber:
@labkey-jeckels and @ankurjuneja: I had EHR open for a different PR and thought I would add this. In #687, there is code for a new mechanism to registerLabworkOverrides(). As I commented there, is that mechanism really needed? Other similar EHR resources are resolved in module dependency order, meaning that rather than needing to maintain separate lists for 'normal' and 'overrides', we might be able to just have one list. If a center module registers a resource with the same name as a built-in module, it would be preferentially used in favor of the built-in one. While LabworkTypes does not currently work this way, it is a simple fix to make it do that.

It is possible there is some complex use case I dont understand that requires a separate method. However, since a LabworkType already stores the owning module and implements isEnabled(), it seems unnecessary for registerLabworkOverride() to require a separate Module argument. If there is some reason that a caller would need a different owning module for the LabworkType and when registerLabworkOverride(), then that caller might be able to override isEnabled() on the LabworkType. If the use-case involved trying to disable a built-in LabworkType with a non-functional one (i.e. just remove the built-in), then I think one could register a LabworkType of the same name and override getResults() to return nothing. This NoOpLabworkType would replace the built-in and never report anything.

This PR leaves the existing registerLabworkOverrides, but deprecates it, so it's backwards compatible. However, if the LabworkType being registered uses the same Module and owningModule, then the code just registers that LabworkType with the regular entrypoint, since using an override is probably not needed.

I'm making assumptions about the motivations behind registerLabworkOverrides(), but if those assumption are valid then this PR should simplify things and make LabworkTypes more consistent with other EHR resources. 

Also, this is targeting 23.11, but I am not especially concerned about the target branch. Develop is also fine.

### DIFF
--- a/ehr/api-src/org/labkey/api/ehr/EHRService.java
+++ b/ehr/api-src/org/labkey/api/ehr/EHRService.java
@@ -329,14 +329,6 @@ abstract public class EHRService
     abstract public List<String> ensureStudyQCStates(Container c, final User u, final boolean commitChanges);
 
     /**
-     * Note: there is probably no need to use this method. Modules can simply register LabworkTypes using registerLabworkType().
-     * LabworkTypes are resolved in module dependency order, meaning if a center-specific module registers a type with the same name as a built-in
-     * type, the LabworkType from the center-specific module will be preferentially used instead of the built-in one.
-     */
-    @Deprecated
-    abstract public void registerLabWorkOverrides(Module module, String fromType, LabworkType toType);
-
-    /**
      * The EHR has a built-in GeneticsCalculations pipeline job that computes inbreeding and kinship based on the pedigree.
      * These are normally calculated in R, saved as TSVs, and imported using java code. This method is a separate entrypoint
      * that allows other code perform the calculations, save the results as TSVs, and then trigger import here.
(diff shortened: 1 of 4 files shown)


ankurjuneja:
Thanks @bbimber for working on this, I lost track of the related PR and forgot about it. I'll test it locally and inform you if I encounter any issues.

ankurjuneja (approved):
This works for me locally and I think we do not need to support the deprecated registerLabWorkOverrides as I have updated JHU code to not use this method and I don't think it affects any other centers, so the deprecated code can be safely removed.

bbimber:
@ankurjuneja i cannot see the JHU repo, but is that staged in a branch of the same name? i could remove that code here, but i assume it'll break JHU unless it's coordinated

ankurjuneja:
Yes, the JHU branch name is same. The code can be removed safely and if you don't mind, I can merge both PRs together once the JHU one gets approved.

bbimber:
OK. done

PR #739, opened 2024-03-15
Use PreparedStatement instead of looping Table.insert

bbimber:
@labkey-martyp, this is tangential to #730. Now that the kinship process is actually working, I forgot how long the insert takes. Matt keyed me in to using PreparedStatement instead of looping over Table.insert(). This should be a minor change that improves insert perf.

### DIFF
--- a/ehr/resources/pipeline/kinship/populateKinship.r
+++ b/ehr/resources/pipeline/kinship/populateKinship.r
@@ -68,4 +68,5 @@ for (species in unique(allPed$Species)){
 
 # write TSV to disk
 print(paste0('Total kinship records: ', nrow(newRecords)))
+newRecords <- dplyr::arrange(newRecords, Id, Id2)
 write.table(newRecords, file = "kinship.txt", append = FALSE, row.names = FALSE, quote = FALSE, sep = '\t')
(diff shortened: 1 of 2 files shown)


labkey-martyp (approved):
This is great. Thanks.

bbimber:
hi @labkey-martyp, I just added one additional small change. this commit allows the job to monitor for being cancelled, and respect that. you already approved, but given this is a change I thought I'd ping you on it.

Also, I added a sort on the TSV produced by the kinship R script. This should have no functional change with import now, but it occurred to me that given how rarely these data change, it is probably unnecessary to constantly re-import everything. If PreparedStatement isnt enough of a speed boost, the other option would be to read two streams: one from the TSV and one from the kinship table with existing records. If those are sorted on Id/Id2, it would be pretty easy to walk those and only do the import/update/delete for records that have actually changed.

bbimber:
hey @labkey-martyp: can you confirm you're still good with this? you may not have seen, but I added a couple minor changes since the first approval. github will let me merge but doesnt seem in the spirit of review. note: these test failures to not seem related and the tests to exercise kinship seem to have passed.

labkey-martyp:
@bbimber Your changes look good. Thanks. Yeah looks like those are just flaky test failures not related to this PR.

PR #743, opened 2024-03-21
Ticket 49626: Performance issues with blood requests

labkey-martyp:
#### Rationale
Blood requests uses study.StudyData to process a request that could span multiple datasets. This was causing performance issues due to study.StudyData not using index scans when looking up columns that don't exist on all datasets (requestid). Adding/altering indexes to speed this up was investigated but postgres continues to not use index scans in this scenario. This PR factors out the use of study.StudyData for the processing of requests. Instead using a union of all datasets that contain requestids.

#### Changes
* Find all datasets with requestid column
* Create union of all datasets with requestid column
* Query this sql function instead of study.StudyData

### DIFF
--- a/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
+++ b/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
@@ -46,6 +46,7 @@ import org.labkey.api.ehr.EHRDemographicsService;
 import org.labkey.api.ehr.EHRQCState;
 import org.labkey.api.ehr.EHRService;
 import org.labkey.api.ehr.demographics.AnimalRecord;
+import org.labkey.api.exp.api.StorageProvisioner;
 import org.labkey.api.ldk.notification.NotificationService;
 import org.labkey.api.query.BatchValidationException;
 import org.labkey.api.query.DuplicateKeyException;
@@ -1458,14 +1459,61 @@ public class TriggerScriptHelper
         }
     }
 
+    // Get all datasets that have a requestid column
+    private List<TableInfo> getRequestDatasets(Container c, User u)
+    {
+        Study study = StudyService.get().getStudy(c);
+        if (null == study)
+        {
+            throw new IllegalStateException("Study not found for container: " + c.getName());
+        }
+
+        List<? extends Dataset> datasets = study.getDatasets();
+        List<TableInfo> requestDatasets = new ArrayList<>();
+        for (Dataset ds : datasets)
+        {
+            TableInfo ti = StorageProvisioner.createTableInfo(ds.getDomain());  // This is faster than ds.getTableInfo(u)
+            if (ti.getColumn("requestid") != null)
+            {
+                requestDatasets.add(ti);
+            }
+        }
+
+        return requestDatasets;
+    }
+
+    // Creates SQL comparable to study.StudyData but only includes datasets with a requestid column and queries qcstate only
+    private SQLFragment getRequestStudyDataSQL(Container c, User u, String requestId)
+    {
+        List<TableInfo> requestDatasets = getRequestDatasets(c, u);
+
+        SQLFragment unionSql = new SQLFragment();
+        for (TableInfo ti : requestDatasets)
+        {
+            if (!unionSql.isEmpty())
+            {
+                unionSql.append(" UNION ALL \n");
+            }
+            else
+            {
+                unionSql.append("\n(");
+            }
+            unionSql.append("SELECT requestid, qcstate FROM ").append(ti);
+        }
+
+        SQLFragment sql = new SQLFragment("SELECT sd.qcstate FROM ");
+
+        return sql.append(unionSql).append(") sd \n").append( "WHERE sd.requestid = ?").add(requestId);
+    }
+
     public void processModifiedRequests(final List<String> requestIds)
     {
+
         _log.info("processing request status for " + requestIds.size() + " records");
         for (String requestId : requestIds)
         {
-            TableInfo studyDataTable = getTableInfo("study", "StudyData");
-            SimpleFilter filter = new SimpleFilter(FieldKey.fromString("requestid"), requestId, CompareType.EQUAL);
-            TableSelector selector = new TableSelector(studyDataTable, Collections.singleton("qcstate"), filter, null);
+            SQLFragment sql = getRequestStudyDataSQL(getContainer(), getUser(), requestId);
+            SqlSelector selector = new SqlSelector(StudyService.get().getDatasetSchema(), sql);
 
             // set the parent ehr.requests record to a specific QC State if all rows for that request match
             Integer requestState = null;


ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -1458,14 +1459,61 @@ public void updateParentRequestQcState(String requestId, Integer qcStateId)
         }
     }
 
+    // Get all datasets that have a requestid column
+    private List<TableInfo> getRequestDatasets(Container c, User u)
```
bbimber, line 1463: Hi @labkey-martyp: this doesnt surprise me that studydata is very slow. Two comments:

- I dont know how to do it better, but I believe at minimum ONPRC (and maybe other centers) have taskId, requestId, etc., added to all/most datasets. EHRManager.ensureDatasetPropertyDescriptors() does this. Even if all datasets have the property for some centers, separating this query probably makes sense.
- Regardless of the answer above, would it make sense to loop datasets and execute the SQL and update once per table, rather that making any kind of union SQL? Each query should be able to use the requestId index.

bbimber:
@labkey-martyp: along the lines of my comment, this branch identifies a list of TableInfos and just executes one query/each, rather than bothering to build any UNION sql. Especially if the set of tables with a requestId column remains large, would this get at the problem better? https://github.com/LabKey/ehrModules/compare/release23.11-SNAPSHOT...23.11_fb_trigger_refactor2

Note: this diff is rendering weird, but the changes are relatively minor.

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -1458,14 +1459,61 @@ public void updateParentRequestQcState(String requestId, Integer qcStateId)
...
+            throw new IllegalStateException("Study not found for container: " + c.getName());
+        }
+
+        List<? extends Dataset> datasets = study.getDatasets();
+        List<TableInfo> requestDatasets = new ArrayList<>();
+        for (Dataset ds : datasets)
+        {
+            TableInfo ti = StorageProvisioner.createTableInfo(ds.getDomain());  // This is faster than ds.getTableInfo(u)
+            if (ti.getColumn("requestid") != null)
+            {
```
labkey-jeckels, line 1477: Even faster to avoid the `TableInfo` when we don't need it.

```suggestion
            if (ds.getDomain().getPropertyByName("requestid") != null)
            {
                TableInfo ti = StorageProvisioner.createTableInfo(ds.getDomain());  // This is faster than ds.getTableInfo(u)
```

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -1458,14 +1459,61 @@ public void updateParentRequestQcState(String requestId, Integer qcStateId)
...
+                unionSql.append(" UNION ALL \n");
+            }
+            else
+            {
+                unionSql.append("\n(");
+            }
+            unionSql.append("SELECT requestid, qcstate FROM ").append(ti);
+        }
+
+        SQLFragment sql = new SQLFragment("SELECT sd.qcstate FROM ");
```
labkey-jeckels, line 1504:
```suggestion
        SQLFragment sql = new SQLFragment("SELECT DISTINCT sd.qcstate FROM ");
```

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
labkey-jeckels: I'm curious on whether there's a perf difference between this FB and Ben's. In general it should be somewhat faster to do one query instead of dozens, but it all comes down to what query plan the DB chooses.

labkey-jeckels (approved):
Approving on the assumption this is at least as fast as issuing separate queries like in the alternative FB.

bbimber:
Couple points:
- I have no doubt that anything involving studydata is probably going to be incredibly slow involving EHR
- The assumption that having requestId column is a heuristic for whether a dataset is involved in requests isnt really built into the assumptions of the EHR module's code. I only see ONPRC EHR using this (i cant read most center-specific EHR modules), but there is an existing core EHR action to ensure 100% of datasets contain a requstId column: https://github.com/LabKey/onprcEHRModules/blob/324d8f1b7102d90b5f99ff3fb4302f0c081b2043/onprc_ehr/test/src/org/labkey/test/tests/onprc_ehr/AbstractGenericONPRC_EHRTest.java#L220. If all the datasets have a requestId column anyway, the auto-generated SQL of this PR and the built-in studydata should not actually be that different. I understand this might not be the status quo on other centers. 
- I dont know how often these pieces of code are encountered, but there are definitely pieces in EHR that assume all datasets include the columns added in ensureDatasetPropertyDescriptors(). 

It might be reasonable to introduce the idea into EHR that having a requestId column should be an opt-in for request behavior, but that's not really the case today.

labkey-martyp:
> Couple points:
> 
> * I have no doubt that anything involving studydata is probably going to be incredibly slow involving EHR
> * The assumption that having requestId column is a heuristic for whether a dataset is involved in requests isnt really built into the assumptions of the EHR module's code. I only see ONPRC EHR using this (i cant read most center-specific EHR modules), but there is an existing core EHR action to ensure 100% of datasets contain a requstId column: https://github.com/LabKey/onprcEHRModules/blob/324d8f1b7102d90b5f99ff3fb4302f0c081b2043/onprc_ehr/test/src/org/labkey/test/tests/onprc_ehr/AbstractGenericONPRC_EHRTest.java#L220. If all the datasets have a requestId column anyway, the auto-generated SQL of this PR and the built-in studydata should not actually be that different. I understand this might not be the status quo on other centers.
> * I dont know how often these pieces of code are encountered, but there are definitely pieces in EHR that assume all datasets include the columns added in ensureDatasetPropertyDescriptors().
> 
> It might be reasonable to introduce the idea into EHR that having a requestId column should be an opt-in for request behavior, but that's not really the case today.

This fix is for WNPRC. About 2/3 of their datasets contain requestid. Their server may pre-date ensureDatasetPropertyDescriptors, I'm not sure. Other centers aren't using requests very much so likely only a few datasets (if any) with requestid is needed for them. We've gone toward using a shared config in datasets_metadata for ensuring the necessary EHR columns like taskid, objectid, remark, etc. That way we have all the column definitions in one place. Since centers aren't using requests as much (if at all), we have not been implementing requestid broadly across all datasets.

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
labkey-martyp: Added

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
labkey-martyp: Added

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
labkey-martyp: I tested both out and they were comparable in perf. Both a big improvement over using StudyData. A single union query was a little faster on the db side so I stuck with that.

PR #747, opened 2024-03-29
Migrate inline javascript event handlers

ankurjuneja:
#### Rationale
A strict (no 'unsafe-inline' directive) CSP forbids all inline events. Instead, events must be attached via JavaScript inside a properly nonced script tag

#### Changes
* snapshotPanel.js
* DefaultEHRCustomizer
* VetReviewDisplayColumn

### DIFF
--- a/ehr/api-src/org/labkey/api/ehr/table/VetReviewDisplayColumn.java
+++ b/ehr/api-src/org/labkey/api/ehr/table/VetReviewDisplayColumn.java
@@ -44,9 +44,10 @@ public class VetReviewDisplayColumn extends DataColumn
                     text = text.replaceAll("\\*\\*", "<span style=\"background-color: yellow;\">\\*\\*</span>");
                 }
 
-                out.write("<a style=\"max-width: 500px;\" onclick=\"EHR.panel.ClinicalManagementPanel.replaceSoap({objectid: " + PageFlowUtil.jsString(StringUtils.trimToNull(tokens[2])) + ", scope: this, callback: function(){EHR.panel.ClinicalManagementPanel.updateVetColumn(this, arguments[0], arguments[1]);}});\">");
-                out.write(text);
-                out.write("</a>");
+                Link.LinkBuilder link = new Link.LinkBuilder(text)
+                        .style("max-width: 500px;")
+                        .onClick("EHR.panel.ClinicalManagementPanel.replaceSoap({objectid: " + PageFlowUtil.jsString(StringUtils.trimToNull(tokens[2])) + ", scope: this, callback: function(){EHR.panel.ClinicalManagementPanel.updateVetColumn(this, arguments[0], arguments[1]);}});");
+                link.appendTo(out);
             }
         }
     }
--- a/ehr/api-src/org/labkey/api/ehr/table/VetReviewDisplayColumn.java
+++ b/ehr/api-src/org/labkey/api/ehr/table/VetReviewDisplayColumn.java
@@ -6,6 +6,7 @@ import org.labkey.api.data.ColumnInfo;
 import org.labkey.api.data.DataColumn;
 import org.labkey.api.data.RenderContext;
 import org.labkey.api.util.PageFlowUtil;
+import org.labkey.api.view.HttpView;
 import org.labkey.api.view.template.ClientDependency;
 
 import java.io.IOException;
@@ -15,6 +16,8 @@ import java.util.Set;
 
 public class VetReviewDisplayColumn extends DataColumn
 {
+    private boolean _clickHandlerRegistered = false;
+
     public VetReviewDisplayColumn(ColumnInfo col)
     {
         super(col);
@@ -44,10 +47,15 @@ public class VetReviewDisplayColumn extends DataColumn
                     text = text.replaceAll("\\*\\*", "<span style=\"background-color: yellow;\">\\*\\*</span>");
                 }
 
-                Link.LinkBuilder link = new Link.LinkBuilder(text)
-                        .style("max-width: 500px;")
-                        .onClick("EHR.panel.ClinicalManagementPanel.replaceSoap({objectid: " + PageFlowUtil.jsString(StringUtils.trimToNull(tokens[2])) + ", scope: this, callback: function(){EHR.panel.ClinicalManagementPanel.updateVetColumn(this, arguments[0], arguments[1]);}});");
-                link.appendTo(out);
+                out.write("<a style=\"max-width: 500px;\" class=\"labkey-text-link vrdc-row\" data-objectid=" + PageFlowUtil.jsString(StringUtils.trimToNull(tokens[2])) + ">");
+
+                if (!_clickHandlerRegistered)
+                {
+                    HttpView.currentPageConfig().addHandlerForQuerySelector("a.vrdc-row", "click", "EHR.panel.ClinicalManagementPanel.replaceSoap({objectid: this.attributes.getNamedItem('data-objectid').value, scope: this, callback: function(){EHR.panel.ClinicalManagementPanel.updateVetColumn(this, arguments[0], arguments[1]);}});" );
+                    _clickHandlerRegistered = true;
+                }
+                out.write(text);
+                out.write("</a>");
             }
         }
     }
--- a/ehr/api-src/org/labkey/api/ehr/table/VetReviewDisplayColumn.java
+++ b/ehr/api-src/org/labkey/api/ehr/table/VetReviewDisplayColumn.java
@@ -47,7 +47,7 @@ public class VetReviewDisplayColumn extends DataColumn
                     text = text.replaceAll("\\*\\*", "<span style=\"background-color: yellow;\">\\*\\*</span>");
                 }
 
-                out.write("<a style=\"max-width: 500px;\" class=\"labkey-text-link vrdc-row\" data-objectid=" + PageFlowUtil.jsString(StringUtils.trimToNull(tokens[2])) + ">");
+                out.write("<a style=\"max-width: 500px;\" class=\"labkey-text-link vrdc-row\" data-objectid=\"" + PageFlowUtil.filter(StringUtils.trimToNull(tokens[2])) + "\">");
 
                 if (!_clickHandlerRegistered)
                 {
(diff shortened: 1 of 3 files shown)


ehr/api-src/org/labkey/api/ehr/table/VetReviewDisplayColumn.java
```diff
@@ -44,9 +44,10 @@ public void renderGridCellContents(RenderContext ctx, Writer out) throws IOExcep
                     text = text.replaceAll("\\*\\*", "<span style=\"background-color: yellow;\">\\*\\*</span>");
                 }
 
-                out.write("<a style=\"max-width: 500px;\" onclick=\"EHR.panel.ClinicalManagementPanel.replaceSoap({objectid: " + PageFlowUtil.jsString(StringUtils.trimToNull(tokens[2])) + ", scope: this, callback: function(){EHR.panel.ClinicalManagementPanel.updateVetColumn(this, arguments[0], arguments[1]);}});\">");
-                out.write(text);
-                out.write("</a>");
+                Link.LinkBuilder link = new Link.LinkBuilder(text)
```
bbimber: Unless I am misunderstanding how LinkBuilder works, this is going to render a totally unique handler per row in the dataframe, right? 

To avoid that, PageConfig.addHandlerForQuerySelector() exists. I recently migrated RequestScoreActionsDisplayColumnFactory to use that (which is in mgap/BimberLabKeyModules), which has a fairly similar pattern as you're doing here. FlagColumnRenderer is a less on-point example that's in core modules. @labkey-adam is who pointed me toward these.

ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
```diff
@@ -824,7 +824,8 @@ public void renderGridCellContents(RenderContext ctx, Writer out) throws IOExcep
                         Date date = (Date) ctx.get("date");
                         Object id = ctx.get(ID_COL);
 
-                        out.write("<span style=\"white-space:nowrap\"><a href=\"javascript:void(0);\" onclick=\"EHR.window.ClinicalHistoryWindow.showClinicalHistory('" + objectid + "', '" + id + "', '" + date + "', this);\">[Show Hx]</a></span>");
+                        Link.LinkBuilder link = new Link.LinkBuilder("[Show Hx]").onClick("EHR.window.ClinicalHistoryWindow.showClinicalHistory('" + objectid + "', '" + id + "', '" + date + "', this);");
+                        DOM.createHtmlFragment(SPAN(at(style, "white-space:nowrap"), link, null)).appendTo(out);
```
bbimber: same comment as above

bbimber (changes requested):

ehr/api-src/org/labkey/api/ehr/table/VetReviewDisplayColumn.java
labkey-adam: Thanks for noticing & commenting, @bbimber

ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
```diff
@@ -824,7 +827,12 @@ public void renderGridCellContents(RenderContext ctx, Writer out) throws IOExcep
                         Date date = (Date) ctx.get("date");
                         Object id = ctx.get(ID_COL);
 
-                        out.write("<span style=\"white-space:nowrap\"><a href=\"javascript:void(0);\" onclick=\"EHR.window.ClinicalHistoryWindow.showClinicalHistory('" + objectid + "', '" + id + "', '" + date + "', this);\">[Show Hx]</a></span>");
+                        out.write("<span style=\"white-space:nowrap\"><a class=\"labkey-text-link anm-history\">[Show Hx]</a></span>");
```
bbimber: Isnt this going to use the same object ID for every row?

bbimber (changes requested):

ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
```diff
@@ -827,10 +827,10 @@ public void renderGridCellContents(RenderContext ctx, Writer out) throws IOExcep
                         Date date = (Date) ctx.get("date");
                         Object id = ctx.get(ID_COL);
 
-                        out.write("<span style=\"white-space:nowrap\"><a class=\"labkey-text-link anm-history\">[Show Hx]</a></span>");
+                        out.write("<span style=\"white-space:nowrap\"><a class=\"labkey-text-link anm-history\" data-objectid=" + objectid + " data-id=" + id + " data-date=" + date.toString() + ">[Show Hx]</a></span>");
```
bbimber: @ankurjuneja: two things:

1) Shouldnt all of the data-* attributes get quoted (line 830)? I think wrapping in jsString() would do it. 

2) This predates your PR, as far as I can tell (and I dont have every single EHR center-specific module locally), both date and objectId are always null when showClinicalHistory() is called, and they are not used for anything in that method. 

As a side point, and this also pre-dates your change, date-serialization into strings using toString() without strictly controlling format and handling of GMT offsets can give many problematic fringe cases, especially on the borders of days. That is, if one side serializes the date assuming the local time zone, but the user side parses it assuming GMT, you can get dates that shift by entire days. 

Unless there is a good reason to actually serialize them, and I dont see one based on the modules I have locally, I suggest not serializing either objectId or date here, and passing null to EHR.window.ClinicalHistoryWindow.showClinicalHistory() for both arguments. That's marginally less text written to the page and avoids date-serialization. It's also in line with every other usage I see. 

This does suggest changing the signature of showClinicalHistory() would be a good idea, but that would touch more code, and is probably out of scope for this.

bbimber (changes requested):

ehr/api-src/org/labkey/api/ehr/table/VetReviewDisplayColumn.java
```diff
@@ -44,7 +47,13 @@ public void renderGridCellContents(RenderContext ctx, Writer out) throws IOExcep
                     text = text.replaceAll("\\*\\*", "<span style=\"background-color: yellow;\">\\*\\*</span>");
                 }
 
-                out.write("<a style=\"max-width: 500px;\" onclick=\"EHR.panel.ClinicalManagementPanel.replaceSoap({objectid: " + PageFlowUtil.jsString(StringUtils.trimToNull(tokens[2])) + ", scope: this, callback: function(){EHR.panel.ClinicalManagementPanel.updateVetColumn(this, arguments[0], arguments[1]);}});\">");
+                out.write("<a style=\"max-width: 500px;\" class=\"labkey-text-link vrdc-row\" data-objectid=" + PageFlowUtil.jsString(StringUtils.trimToNull(tokens[2])) + ">");
```
labkey-jeckels: jsString won't handle all possible values
```suggestion
                out.write("<a style=\"max-width: 500px;\" class=\"labkey-text-link vrdc-row\" data-objectid=\"" + PageFlowUtil.filter(StringUtils.trimToNull(tokens[2])) + "\">");
```

ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
```diff
@@ -824,7 +827,12 @@ public void renderGridCellContents(RenderContext ctx, Writer out) throws IOExcep
                         Date date = (Date) ctx.get("date");
                         Object id = ctx.get(ID_COL);
 
-                        out.write("<span style=\"white-space:nowrap\"><a href=\"javascript:void(0);\" onclick=\"EHR.window.ClinicalHistoryWindow.showClinicalHistory('" + objectid + "', '" + id + "', '" + date + "', this);\">[Show Hx]</a></span>");
+                        out.write("<span style=\"white-space:nowrap\"><a class=\"labkey-text-link anm-history\" data-objectid=" + PageFlowUtil.filter(objectid) + " data-id=" + PageFlowUtil.filter(id) + " data-date=" + PageFlowUtil.filter(date) + ">[Show Hx]</a></span>");
```
labkey-jeckels: Maybe these values will never have spaces, but they'd break the HTML if they do.
```suggestion
                        out.write("<span style=\"white-space:nowrap\"><a class=\"labkey-text-link anm-history\" data-objectid=\"" + PageFlowUtil.filter(objectid) + "\" data-id=\"" + PageFlowUtil.filter(id) + "\" data-date=\"" + PageFlowUtil.filter(date) + "\">[Show Hx]</a></span>");
```

labkey-jeckels (approved):
Some encoding adjustments but otherwise looks OK to me.

I suggest handling any date changes separately. Valid concern, but unrelated to CSP.

@labkey-jeckels : i think i was proposing a reasonable middle ground. again, while the concern pre-dates this PR, blindly serializing a date to string is a bad idea. Fortunately, all we need to do here is pass NULL and dont serialize anything. This also results in not needing to serialize objectId and date on into data- attributes for every table row. Neither value is used for anything within EHR.window.ClinicalHistoryWindow.showClinicalHistory(). 

Like I said above, I completely agree changing EHR.window.ClinicalHistoryWindow.showClinicalHistory() itself is probably out of scope for this PR.

ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
ankurjuneja: hi @bbimber , that is a good suggestion. Would be great to followup on the second point later. Can you please review again?

ehr/src/org/labkey/ehr/table/DefaultEHRCustomizer.java
```diff
@@ -824,7 +827,12 @@ public void renderGridCellContents(RenderContext ctx, Writer out) throws IOExcep
                         Date date = (Date) ctx.get("date");
                         Object id = ctx.get(ID_COL);
 
-                        out.write("<span style=\"white-space:nowrap\"><a href=\"javascript:void(0);\" onclick=\"EHR.window.ClinicalHistoryWindow.showClinicalHistory('" + objectid + "', '" + id + "', '" + date + "', this);\">[Show Hx]</a></span>");
+                        out.write("<span style=\"white-space:nowrap\"><a class=\"labkey-text-link anm-history\" data-objectid=\"" + PageFlowUtil.filter(objectid) + "\" data-id=\"" + PageFlowUtil.filter(id) + "\" data-date=\"" + PageFlowUtil.filter(date) + "\">[Show Hx]</a></span>");
```
bbimber: @ankurjuneja: I really thing you should just: 

1) no longer include data-objectid and data-date at all
2) on line 833 just pass 'null' for those arguments. 

There is no real downside, no change to code outside of what you're already touching here, you fix a potential bug should anyone every use 'date' for anything, and you end up writing less text to the page.

bbimber (changes requested):

bbimber (approved):

bbimber:
thanks @ankurjuneja

PR #756, opened 2024-04-08
Ticket 49214: Expanding the wnprc billing module

labkey-bpatel:
#### Rationale
Currently, the WNPRC Billing module processes only one type of procedure (blood draws) for one type of group (animal care) during a billing run. All the other procedures are handled by manual entries via Misc. Charges.

In this PR, we enhance the Billing module to allow blood draws for more than one group, and also have capability to automate billing for other procedures, such as surgeries, etc. in the future. Note that some effort will still be required in order to capture additional procedures, but that effort should be minimal.

Spec [here](https://docs.google.com/document/d/1EVDGq8_ydHScxmwrZRqXBo4JtoljKl3dO99ENQMnwPA/edit).

#### Related Pull Requests
* https://github.com/LabKey/wnprc-modules/pull/554

#### Changes
* Update ehr_billing schema to add the new table 'procedureQueryChargeIdAssoc' that should have procedures and associated chargeId to be part of the automated billing process.
* Update the billing run process to include procedure charges via the queries listed in 'procedureQueryChargeIdAssoc'
* Update billing notification to have these procedure queries to be included in the charge summary section of the notification.
* Handle table delete upon folder delete, mainly for tests.


ehr_billing/resources/schemas/dbscripts/sqlserver/ehr_billing-23.001-23.002.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE ehr_billing.procedureQueryChargeIdAssoc (
+
+   rowId SERIAL NOT NULL,
+   schemaName varchar(200) NOT NULL,
+   queryName varchar(500) NOT NULL,
+   description varchar(2000) NOT NULL,
```
labkey-jeckels: SQLServer should us `NVARCHAR`

ehr_billing/resources/schemas/dbscripts/sqlserver/ehr_billing-23.001-23.002.sql
```diff
@@ -0,0 +1,19 @@
...
+   chargeId int NOT NULL,
+
+   container ENTITYID NOT NULL,
+   createdBy USERID,
+   created timestamp,
```
labkey-jeckels: SQLServer uses `DATETIME`, not `TIMESTAMP`

ehr_billing/src/org/labkey/ehr_billing/EHR_BillingContainerListener.java
```diff
@@ -93,6 +93,10 @@ public void containerDeleted(Container c, User user)
             if (chargeableItemCategoriesTable.getTableType() == DatabaseTableType.TABLE)
                 Table.delete(chargeableItemCategoriesTable, containerFilter);
 
+            TableInfo procedurequerychargeidassocTable = EHR_BillingSchema.getInstance().getProcedureQueryChargeIdAssoc();
+            if (procedurequerychargeidassocTable.getTableType() == DatabaseTableType.TABLE)
```
labkey-jeckels: I know code for other tables follows this same pattern, but why do we need this check? If the table's not there, we're in a very, very bad state and we don't do similar checks anywhere else.

ehr_billing/src/org/labkey/ehr_billing/pipeline/BillingTask.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2019 LabKey Corporation
+ * Copyright (c) 2024 LabKey Corporation
```
labkey-jeckels:
```suggestion
 * Copyright (c) 2018-2024 LabKey Corporation
```

ehr_billing/src/org/labkey/ehr_billing/pipeline/BillingTask.java
```diff
@@ -452,15 +452,226 @@ private Collection<Map<String, Object>> getRowList(BillingPipelineJobProcess pro
...
+        }
+        getJob().getLogger().info("Finished Caching " + process.getLabel());
+    }
+
+    private ChargeInfo getChargeInfo(int chargeId, ArrayList<ChargeInfo> chargeInfoArrayList, Object date)
+    {
+        Date chargeDate = (Date) date;
+        List<ChargeInfo> chargeInfoList = chargeInfoArrayList.stream().filter(ci -> ci.getChargeId() == chargeId).toList();
+
+        if (chargeInfoList.size() == 1)
```
labkey-jeckels: Similar feedback to the other PR - no need for these `if` checks. Just the `for` loop is enough

ehr_billing/api-src/org/labkey/api/ehr_billing/notification/BillingNotificationProvider.java
```diff
@@ -87,5 +88,20 @@ public interface BillingNotificationProvider
...
+    /**
+     * @param u User
+     * @param c Container where custom charge categories are defined
+     * @param startDate Filter custom charge category query where date >= startDate
+     * @param endDate Filter custom charge category query where date <= endDate
+     * @param totalsByCategory Add totals for the center-specific custom category to be displayed in the Charge Summary section of the notification.
+     * @param additionalCategoryUrls Add URLs for center-specific custom category to be linked to the Category
+     * See WNPRC Billing's implementation of this method for an example.
+     */
+    public default void getAdditionalChargeCategoryInfo(User u, Container c, Date startDate, Date endDate,
```
labkey-jeckels: It's hard to follow how the two maps are correlated (or not).

Assuming I'm following things correctly, I recommend making this a true `get` method, returning a `List<X>`, where a `X` has a String description, an ActionURL, a double total and a double totalCost.

labkey-jeckels (changes requested):

ehr_billing/src/org/labkey/ehr_billing/EHR_BillingContainerListener.java
labkey-bpatel: good point, now updated

ehr_billing/api-src/org/labkey/api/ehr_billing/notification/BillingNotificationProvider.java
labkey-bpatel: Agreed, its not clear. I've update the definition. I did not return List<X>, but instead return a Map and updated the code where its used, since that was easier to implement that way.

ehr_billing/resources/schemas/dbscripts/sqlserver/ehr_billing-23.001-23.002.sql
labkey-bpatel: good catch, i missed that. thanks

ehr_billing/src/org/labkey/ehr_billing/pipeline/BillingTask.java
labkey-bpatel: yup, now updated

ehr_billing/api-src/org/labkey/api/ehr_billing/notification/BillingNotificationProvider.java
labkey-jeckels: This is much cleaner. But I'd still prefer to have a real Java class with the handful of supported properties instead of a map where we hard-code the property names as strings in multiple places.

ehr_billing/src/org/labkey/ehr_billing/notification/BillingNotification.java
```diff
@@ -165,13 +165,25 @@ public String getMessageBodyHTML(Container c, User u)
...
 
         getProjectSummaryPerCategory(u, startDate, endDate, categoryToQuery, containerMap, dataMap, totalsByCategory, fields);
+        Map<String, Map<String, Object>> additionalCategoryInfo = _notificationProvider.getAdditionalChargeCategoryInfo(u, billingContainer, startDate.getTime(), endDate.getTime());
+
+        for (String category : additionalCategoryInfo.keySet())
+        {
+            totalsByCategory.put(category, new HashMap<>());
+            Map<String, Double> totalsMap = new HashMap<>();
+            Map<String, Object> categoryInfo = additionalCategoryInfo.get(category);
+            totalsMap.put("totalCost", (Double) categoryInfo.get("totalCost"));
```
labkey-jeckels: Should any of these strings use the new methods like `getTotalCostColName()` on `InvoicedItemsProcessingService`?

labkey-jeckels (changes requested):
Let me know if there's a compelling reason to keep using a Map instead of a new class (see my reply on the feedback from the first review)

ehr_billing/api-src/org/labkey/api/ehr_billing/notification/BillingNotificationProvider.java
labkey-bpatel: OK this is now cleaned up further

ehr_billing/src/org/labkey/ehr_billing/notification/BillingNotification.java
labkey-bpatel: This is now refactored as part of the cleanup above

ehr_billing/src/org/labkey/ehr_billing/notification/BillingNotification.java
```diff
@@ -307,34 +313,22 @@ private void getProjectSummary(Container c, User u, final Calendar start, Calend
         final Map<FieldKey, ColumnInfo> cols = QueryService.get().getColumns(ti, fieldKeys);
         TableSelector ts = new TableSelector(ti, cols.values(), null, null);
         ts.setNamedParameters(params);
+        ChargeCategoryInfo categoryInfo = new ChargeCategoryInfo();
+        categoryInfo.setCategoryLabel(categoryName);
+        final double[] totalQuantity = {0.0}; // 1 element final array since 'final' is required in the below exec method
```
labkey-jeckels, line 319: This is fine - no need to change. FYI, org.apache.commons.lang3.mutable.MutableDouble is handy for scenarios too.

ehr_billing/src/org/labkey/ehr_billing/notification/BillingNotification.java
```diff
@@ -410,13 +407,18 @@ protected void createChargeSummaryReport(final StringBuilder msg, Date lastInvoi
...
-            String url = createURL(container, centerSpecificBillingSchema, categoryToQuery.get(category), null) + "&query.param.StartDate=" + getDateFormat(c).format(start.getTime()) + "&query.param.EndDate=" + getDateFormat(c).format(endDate.getTime());
-            msg.append("<tr><td><a href='" + url + "'>" + category + "</a></td><td align='right'>" + _numItemsFormat.format(totalsMap.get("total")) + "</td><td align='right'>" + _dollarFormat.format(totalsMap.get("totalCost")) + "</td></tr>");
+            String url;
+            if (categoryToQuery.containsKey(category))
+                url = createURL(container, centerSpecificBillingSchema, categoryToQuery.get(category), null) + "&query.param.StartDate=" + getDateFormat(c).format(start.getTime()) + "&query.param.EndDate=" + getDateFormat(c).format(endDate.getTime());
+            else
+                url = categoryInfo.getCategoryURL();
+
+            if (null != url)
+                msg.append("<tr><td><a href='" + url + "'>" + category + "</a></td><td align='right'>" + _numItemsFormat.format(categoryInfo.getTotalQuantity()) + "</td><td align='right'>" + _dollarFormat.format(categoryInfo.getTotalCost()) + "</td></tr>");
```
labkey-jeckels: I missed this before. `category` should be HTML-encoded.

labkey-jeckels (approved):
One more request for HTML encoding but no need for further review.

ehr_billing/src/org/labkey/ehr_billing/notification/BillingNotification.java
labkey-adam: `url` and even formatted values should be HTML encoded as well. It's easy to set format patterns that produce HTML-unsafe results.

labkey-bpatel: Ah yup! thank you, both!

ehr_billing/src/org/labkey/ehr_billing/notification/BillingNotification.java
labkey-bpatel: I see, that's good to know! thank you

PR #761, opened 2024-04-15
Add a 24.00x db script to create a new ehr_billing table if it doesn't exist

labkey-bpatel:
#### Rationale
Erring on the side caution, add 24.00x db script to create a new table (ehr_billing.procedureQueryChargeIdAssoc) if it doesn't already exist since the module version is 24.00x, and the new table was created in 23.11 branch with a 23.00x script.

#### Related Pull Requests
* 

#### Changes
*


labkey-bpatel:
Opened a 23.4 PR instead of 24.3 🙄

PR #762, opened 2024-04-15
Add a 24.00x db script to create a new ehr_billing table if it doesn't exist 

labkey-bpatel:
#### Rationale
Erring on the side caution, adding a 24.00x db script to create a new table (ehr_billing.procedureQueryChargeIdAssoc) if it doesn't already exist since the module version here is already bumped to 24.00x, and the new table was created in 23.11 branch with a 23.00x script.

#### Changes
* Upgrade scripts to create new ehr_billing.procedureQueryChargeIdAssoc table if it doesn't exist for 24.3 servers


labkey-martyp (approved):

labkey-adam (approved):
This is great, thanks for catching and taking care of it. Should we back-port this schema version bump and the others in ehrModules to 23.11? That would ensure that all future scripts are versioned correctly for 23.11, 24.3, develop, and future releases.

labkey-bpatel:
> This is great, thanks for catching and taking care of it. Should we back-port this schema version bump and the others in ehrModules to 23.11? That would ensure that all future scripts are versioned correctly for 23.11, 24.3, develop, and future releases.

Good idea, will open a backport PR shortly here

PR #764, opened 2024-04-16
Backport schema version from 24.3 for various EHR modules

labkey-bpatel:
#### Rationale
Backport https://github.com/LabKey/ehrModules/pull/762 to ensure future scripts are versioned correctly.

#### Changes
* Backport changes


labkey-adam (changes requested):
I don't think there's any value in back-porting the scripts themselves... I think we just need to back-port the schema version updates in all modules where there's any chance that new scripts will appear in 23.11. Clearly `EHR_BillingModule.java` and `EHRModule.java`, where we've already seen scripts. Maybe there are other modules that should be bumped (I haven't inspected all of them).

labkey-adam (approved):

PR #771, opened 2024-04-25
Add optional check for blood draws nearing the max vol

csebranek:
#### Rationale
WNPRC wants to warn users when filling out a blood draw in EHR, if an animal is getting close to an its max allowable blood draw limit. Currently EHR does this through a TriggerScriptHelper that has a lot of logic to check for blood remaining, but only warns if an animal goes over the limit. In order to take advantage of that work that is already done in the TriggerScriptHelper, it seemed most reliable to tap into this by adding in an optional limit check that can be turned on in a trigger script. This change is backwards compatible with other areas, and requires no intervention for other centers.

#### Changes
* TriggerScriptHelper.java - added on an error msg to check additional limits, made it opt-in behavior. Added member to BloodInfo to track report only records in the current transaction.

* species.tsv - needed to test validation for blood volumes

* labwork_services.tsv - does not really apply here, but wanted to get started on this. blood.js in EHR creates a clinpath request after a request becomes public, there is no EHR test for this. This puts in the minimum required data if WNPRC wants to test it later.

### DIFF
--- a/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
+++ b/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
@@ -1258,13 +1258,12 @@ public class TriggerScriptHelper
                                .append(" kg).\n");
             }
 
-            //report all the ones that are close to limit
-            for (double closeValue : closeToThreshold.descendingSet())
+            if (!closeToThreshold.isEmpty())
             {
                 errorMsgBuilder.append("Limit notice! Blood volume of ")
                                .append(rowQuantity)
                                .append(" (")
-                               .append(closeValue)
+                               .append(closeToThreshold.descendingSet().iterator().next())
                                .append(" over ")
                                .append(interval)
                                .append(" days) is within ")
@@ -1285,10 +1284,15 @@ public class TriggerScriptHelper
     }
 
     /**
-     * Gets the center specific threshold to warn when blood vols are close to the max blood allowed,
+     * Gets the center specific threshold from _centerCustomProps.bloodNearOverageThreshold to warn when blood vols are close to the max blood allowed,
      * only used if the doWarnForBloodNearOverages() is true
      * e.g., if the max allowable blood drawn vol is 60.0, a threshold of 4.0 will warn users if blood vol is greater than 56.0 ml
-     * this should be set on the JS side with helper.setCenterCustomProps()
+     * this should be set in a JS trigger script via     helper.setCenterCustomProps(), for example:
+     * helper.setCenterCustomProps({
+     *  doWarnForBloodNearOverages: true,
+     *  bloodNearOverageThreshold: 5.0
+     * })
+     * It uses default value "_bloodNearingOveragesThresholdDefaultValue" if none is supplied or incorrect data type is supplied
      *
      * @return      the threshold value of the limit
      */
@@ -1306,6 +1310,10 @@ public class TriggerScriptHelper
             {
                 return (Double) theVal;
             }
+            else
+            {
+                _log.warn("TriggerScriptHelper.getBloodNearingOveragesThreshold incorrect datatype supplied for _centerCustomProps.bloodNearOverageThreshold, using default value.");
+            }
         }
         else
         {
@@ -1317,8 +1325,12 @@ public class TriggerScriptHelper
 
 
     /**
-     * For use with _bloodLimitThreshold, checks to see whether we should warn about bloods draws nearing their limit,
-     * this should be set on the JS side with helper.setCenterCustomProps()
+     * For use with getBloodNearingOveragesThreshold(), checks to see whether we should warn about bloods draws nearing their limit,
+     * this should be set in a JS trigger script via helper.setCenterCustomProps(), for example:
+     * helper.setCenterCustomProps({
+     *  doWarnForBloodNearOverages: true,
+     *  bloodNearOverageThreshold: 5.0
+     * })
      *
      * @return      whether to warn for bloods nearing overages
      */
--- a/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
+++ b/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
@@ -1307,6 +1307,10 @@ public class TriggerScriptHelper
                 return (Double) theVal;
             }
         }
+        else
+        {
+            _log.warn("TriggerScriptHelper.getBloodNearingOveragesThreshold no value found for _centerCustomProps.bloodNearOverageThreshold, using default value.");
+        }
         return _bloodNearingOveragesThresholdDefaultValue;
     }
 
--- a/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
+++ b/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
@@ -1860,7 +1860,7 @@ public class TriggerScriptHelper
                 }
                 else if (hasCustomDepartureStatus)
                 {
-                    status = _centerCustomProps.get("departureStatus").toString();
+                    status = String.valueOf(_centerCustomProps.get("departureStatus"));
                 }
                 else
                 {
--- a/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
+++ b/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
@@ -138,6 +138,10 @@ public class TriggerScriptHelper
 
     private static final Logger _log = LogHelper.getLogger(TriggerScriptHelper.class, "Server-side validation of EHR data insert/update/deletes");
 
+    private static boolean _checkThreshold = false;
+
+    private static double _bloodLimitThreshold = 4.0; //in mL
+
     private TriggerScriptHelper(int userId, String containerId)
     {
         User user = UserManager.getUser(userId);
@@ -1142,7 +1146,7 @@ public class TriggerScriptHelper
                             foundRow = true;
                         }
                     }
-                    BloodInfo bloodsIntransc = new BloodInfo(objectId, ConvertHelper.convert(map.get("date"), Date.class), ConvertHelper.convert(map.get("quantity"), Double.class));
+                    BloodInfo bloodsIntransc = new BloodInfo(objectId, ConvertHelper.convert(map.get("date"), Date.class), ConvertHelper.convert(map.get("quantity"), Double.class), true);
                     allBloods.add(bloodsIntransc);
                 }
                 catch (ConversionException e)
@@ -1186,10 +1190,28 @@ public class TriggerScriptHelper
 
         // Get records from the database in our date range that aren't part of the current transaction
         TableSelector tsdate = new TableSelector(ti, PageFlowUtil.set("objectid", "date", "quantity"), filter, null);
-        allBloods.addAll(tsdate.getArrayList(BloodInfo.class));
+        //get the db object
+            try (Results r = tsdate.getResults())
+            {
+
+                if (tsdate.getRowCount() > 0)
+                {
+                    while (r.next())
+                    {
+                        BloodInfo bi = new BloodInfo(r.getString("objectid"), r.getDate("date"), r.getDouble("quantity"), false);
+                        allBloods.add(bi);
+                    }
+                }
+            }
+            catch (SQLException e)
+            {
+                throw new RuntimeException(e);
+            }
+
 
         // Iterate over all of the blood records
         TreeSet<Double> overages = new TreeSet<>();
+        TreeSet<Double> closeToThreshold = new TreeSet<>();
         for (BloodInfo blood1 : allBloods)
         {
             double bloodNextInterval = 0;
@@ -1215,15 +1237,89 @@ public class TriggerScriptHelper
             {
                 overages.add(bloodNextInterval);
             }
+            else if (getThresholdCheck())
+            {
+                //only report about problematic bloods nearing the limit in the current transaction
+                //because allBloods contains everything and we want to distinguish them
+                if (blood1.getIsInTransaction())
+                {
+                    double maxAllowableThreshold = maxAllowable - getThresholdLimit();
+                    if (bloodNextInterval > maxAllowableThreshold)
+                    {
+                        closeToThreshold.add(bloodNextInterval);
+                    }
+                }
+            }
         }
 
-        //always report the most severe overage
-        if (!overages.isEmpty())
+        if (!overages.isEmpty() || !closeToThreshold.isEmpty())
         {
-            return "Blood volume of " + rowQuantity + " (" + overages.descendingSet().iterator().next() + " over " + interval + " days) exceeds the allowable volume of " + maxAllowable + " mL (weight: " + weight + " kg)";
-        }
+            StringBuilder errorMsgBuilder = new StringBuilder();
 
-        return null;
+            //always report the most severe overage
+            if (!overages.isEmpty())
+            {
+                errorMsgBuilder.append("Blood volume of ");
+                errorMsgBuilder.append(rowQuantity);
+                errorMsgBuilder.append(" (");
+                errorMsgBuilder.append(overages.descendingSet().iterator().next());
+                errorMsgBuilder.append(" over ");
+                errorMsgBuilder.append(interval);
+                errorMsgBuilder.append(" days) exceeds the allowable volume of ");
+                errorMsgBuilder.append(maxAllowable);
+                errorMsgBuilder.append(" mL (weight: ");
+                errorMsgBuilder.append(weight);
+                errorMsgBuilder.append(" kg).\n");
+            }
+
+            //report all the ones that are close to limit
+            for (double closeValue : closeToThreshold.descendingSet())
+            {
+                errorMsgBuilder.append("Limit notice! Blood volume of ");
+                errorMsgBuilder.append(rowQuantity);
+                errorMsgBuilder.append(" (");
+                errorMsgBuilder.append(closeValue);
+                errorMsgBuilder.append(" over ");
+                errorMsgBuilder.append(interval);
+                errorMsgBuilder.append(" days) is within ");
+                errorMsgBuilder.append(getThresholdLimit());
+                errorMsgBuilder.append(" mL of the max allowable limit of ");
+                errorMsgBuilder.append(maxAllowable);
+                errorMsgBuilder.append(" mL (weight: ");
+                errorMsgBuilder.append(weight);
+                errorMsgBuilder.append(" kg).\n");
+            }
+
+            return errorMsgBuilder.toString();
+        }
+        else
+        {
+            return null; // No errors
+        }
+    }
+
+    public double getThresholdLimit()
+    {
+        return _bloodLimitThreshold;
+
+    }
+
+    public void setThresholdLimit(double threshold)
+    {
+        _bloodLimitThreshold = threshold;
+
+    }
+
+    public boolean getThresholdCheck()
+    {
+        return _checkThreshold;
+
+    }
+
+    public void setThresholdCheck(boolean checkIt)
+    {
+        _checkThreshold = checkIt;
+
     }
 
     public static class BloodInfo implements Comparable<BloodInfo>
@@ -1231,14 +1327,16 @@ public class TriggerScriptHelper
         private String _objectId;
         private Date _date;
         private double _quantity;
+        private boolean _inTransaction;
 
         public BloodInfo() {}
 
-        public BloodInfo(String objectId, Date date, Double quantity)
+        public BloodInfo(String objectId, Date date, Double quantity, boolean isInTransaction)
         {
             _objectId = objectId;
             setDate(date);
             _quantity = quantity;
+            _inTransaction = isInTransaction;
         }
 
         @Override
@@ -1276,6 +1374,16 @@ public class TriggerScriptHelper
         public void setQuantity(double quantity)
         {
             _quantity = quantity;
+        }
+
+        public void setIsInTransaction(boolean inTransaction)
+        {
+            _inTransaction = inTransaction;
+        }
+
+        public boolean getIsInTransaction()
+        {
+            return _inTransaction;
         }
 
         final long MILLIS_PER_DAY = 24 * 3600 * 1000;
--- a/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
+++ b/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
@@ -1310,7 +1310,7 @@ public class TriggerScriptHelper
             }
             else
             {
-                throw new RuntimeException("TriggerScriptHelper.getBloodNearingOveragesThreshold invalid value found for _centerCustomProps.bloodNearOverageThreshold.");
+                throw new RuntimeException("TriggerScriptHelper.getBloodNearingOveragesThreshold invalid value found for _centerCustomProps.bloodNearOverageThreshold. Required type is a double.");
             }
         }
         else
--- a/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
+++ b/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
@@ -134,13 +134,11 @@ public class TriggerScriptHelper
      *  <li>departureStatus - center specific custom status for animal departures</li>
      */
     @NotNull
-    private static final Map<String, String> _centerCustomProps = new HashMap<>();
+    private static final Map<String, Object> _centerCustomProps = new HashMap<>();
 
     private static final Logger _log = LogHelper.getLogger(TriggerScriptHelper.class, "Server-side validation of EHR data insert/update/deletes");
 
-    private static boolean _checkThreshold = false;
-
-    private static double _bloodLimitThreshold = 4.0; //in mL
+    private static double _bloodNearingOveragesThresholdDefaultValue = 4.0; //in mL
 
     private TriggerScriptHelper(int userId, String containerId)
     {
@@ -1189,25 +1187,13 @@ public class TriggerScriptHelper
         TableInfo ti = getTableInfo("study", "Blood Draws");
 
         // Get records from the database in our date range that aren't part of the current transaction
-        TableSelector tsdate = new TableSelector(ti, PageFlowUtil.set("objectid", "date", "quantity"), filter, null);
+        TableSelector  bloodQuery = new TableSelector(ti, PageFlowUtil.set("objectid", "date", "quantity"), filter, null);
         //get the db object
-            try (Results r = tsdate.getResults())
-            {
-
-                if (tsdate.getRowCount() > 0)
-                {
-                    while (r.next())
-                    {
-                        BloodInfo bi = new BloodInfo(r.getString("objectid"), r.getDate("date"), r.getDouble("quantity"), false);
-                        allBloods.add(bi);
-                    }
-                }
+        bloodQuery.forEach( rs -> {
+            BloodInfo bi = new BloodInfo(rs.getString("objectid"), rs.getDate("date"), rs.getDouble("quantity"), false);
+            allBloods.add(bi);
             }
-            catch (SQLException e)
-            {
-                throw new RuntimeException(e);
-            }
-
+        );
 
         // Iterate over all of the blood records
         TreeSet<Double> overages = new TreeSet<>();
@@ -1237,13 +1223,13 @@ public class TriggerScriptHelper
             {
                 overages.add(bloodNextInterval);
             }
-            else if (getThresholdCheck())
+            else if (doWarnForBloodNearOverages())
             {
                 //only report about problematic bloods nearing the limit in the current transaction
                 //because allBloods contains everything and we want to distinguish them
-                if (blood1.getIsInTransaction())
+                if (blood1.getInTransaction())
                 {
-                    double maxAllowableThreshold = maxAllowable - getThresholdLimit();
+                    double maxAllowableThreshold = maxAllowable - getBloodNearingOveragesThreshold();
                     if (bloodNextInterval > maxAllowableThreshold)
                     {
                         closeToThreshold.add(bloodNextInterval);
@@ -1259,35 +1245,35 @@ public class TriggerScriptHelper
             //always report the most severe overage
             if (!overages.isEmpty())
             {
-                errorMsgBuilder.append("Blood volume of ");
-                errorMsgBuilder.append(rowQuantity);
-                errorMsgBuilder.append(" (");
-                errorMsgBuilder.append(overages.descendingSet().iterator().next());
-                errorMsgBuilder.append(" over ");
-                errorMsgBuilder.append(interval);
-                errorMsgBuilder.append(" days) exceeds the allowable volume of ");
-                errorMsgBuilder.append(maxAllowable);
-                errorMsgBuilder.append(" mL (weight: ");
-                errorMsgBuilder.append(weight);
-                errorMsgBuilder.append(" kg).\n");
+                errorMsgBuilder.append("Blood volume of ")
+                               .append(rowQuantity)
+                               .append(" (")
+                               .append(overages.descendingSet().iterator().next())
+                               .append(" over ")
+                               .append(interval)
+                               .append(" days) exceeds the allowable volume of ")
+                               .append(maxAllowable)
+                               .append(" mL (weight: ")
+                               .append(weight)
+                               .append(" kg).\n");
             }
 
             //report all the ones that are close to limit
             for (double closeValue : closeToThreshold.descendingSet())
             {
-                errorMsgBuilder.append("Limit notice! Blood volume of ");
-                errorMsgBuilder.append(rowQuantity);
-                errorMsgBuilder.append(" (");
-                errorMsgBuilder.append(closeValue);
-                errorMsgBuilder.append(" over ");
-                errorMsgBuilder.append(interval);
-                errorMsgBuilder.append(" days) is within ");
-                errorMsgBuilder.append(getThresholdLimit());
-                errorMsgBuilder.append(" mL of the max allowable limit of ");
-                errorMsgBuilder.append(maxAllowable);
-                errorMsgBuilder.append(" mL (weight: ");
-                errorMsgBuilder.append(weight);
-                errorMsgBuilder.append(" kg).\n");
+                errorMsgBuilder.append("Limit notice! Blood volume of ")
+                               .append(rowQuantity)
+                               .append(" (")
+                               .append(closeValue)
+                               .append(" over ")
+                               .append(interval)
+                               .append(" days) is within ")
+                               .append(getBloodNearingOveragesThreshold())
+                               .append(" mL of the max allowable limit of ")
+                               .append(maxAllowable)
+                               .append(" mL (weight: ")
+                               .append(weight)
+                               .append(" kg).\n");
             }
 
             return errorMsgBuilder.toString();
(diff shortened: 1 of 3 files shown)


bbimber:
@csebranek: thanks for adding me. one high level comment: if we're going to go down this road, it seems like it would be helpful to do whatever refactoring is needed in EHR to let centers implement their center-specific logic. If memory serves, ONPRC already has some degree of custom blood draw logic that overrides what's in core EHR. how much did you look into that here?

csebranek:
@csebranek: I left a few more specific questions/comments. 

While it is somewhat true that the core EHR's JS file should have onInit(), i think there are probably ways we could let a center-specific module get more control, including:
- I bet core EHR could have an onInit() for essential code, and also register a separate handler using EHR.Server.TriggerManager.registerHandlerForQuery(). The second piece could contain the pieces of code a center-specific module needs to override. 
- As noted above, we could refactor a lot of the blood-specific code currently in onInit() into separate functions in a standalone ./scripts/ehr/bloodFunctions.js file. This code could be called from EHR's blood.js like it does now, or a center-specific module could override core EHR and directly call functions defined in this new standalone bloodFunctions.js file. This should be a basic refactor that could potentially give a lot more control to center-specific modules without duplicating code.

This all said, I am not following why overriding init is impossible or even a problem here. I dont think you are intending to override the core EHR's onInit for anything you need. There are lots of examples of center-specific modules that register a custom INIT function, with the purpose of calling setScriptOptions: https://github.com/LabKey/onprcEHRModules/blob/27840ff6a821e4147f62c8c31066d1ef38c71764/onprc_ehr/resources/scripts/onprc_ehr/onprc_triggers.js#L140

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -1142,7 +1146,7 @@ private String checkOtherDrawsQuantity(String id, Date date, String rowObjectId,
                             foundRow = true;
                         }
                     }
-                    BloodInfo bloodsIntransc = new BloodInfo(objectId, ConvertHelper.convert(map.get("date"), Date.class), ConvertHelper.convert(map.get("quantity"), Double.class));
+                    BloodInfo bloodsIntransc = new BloodInfo(objectId, ConvertHelper.convert(map.get("date"), Date.class), ConvertHelper.convert(map.get("quantity"), Double.class), true);
```
bbimber, line 1145: Is 'inTransaction' ever not true? What's it needed for?

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -138,6 +138,10 @@ public class TriggerScriptHelper
 
     private static final Logger _log = LogHelper.getLogger(TriggerScriptHelper.class, "Server-side validation of EHR data insert/update/deletes");
 
+    private static boolean _checkThreshold = false;
+
+    private static double _bloodLimitThreshold = 4.0; //in mL
```
bbimber: It seems like both of these should be set through some mechanism that is easily modified by the modules. The most natural thing would be to add properties to the existing addScriptOption mechanism, which is widely used, like:
```
    helper.setScriptOptions({
        allowAnyId: true,
        requiresStatusRecalc: true,
        allowDeadIds: true,
        skipIdFormatCheck: true,
        skipHousingCheck: true,
        announceAllModifiedParticipants: true
    });
```

This does mean it is set on the JS layer, and would need ot be propagated into the java helper. I would suggest defaulting to false, meaning we preserve the status quo unless a center-specific module opts in.

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -1186,10 +1190,28 @@ private String checkOtherDrawsQuantity(String id, Date date, String rowObjectId,
 
         // Get records from the database in our date range that aren't part of the current transaction
         TableSelector tsdate = new TableSelector(ti, PageFlowUtil.set("objectid", "date", "quantity"), filter, null);
-        allBloods.addAll(tsdate.getArrayList(BloodInfo.class));
+        //get the db object
+            try (Results r = tsdate.getResults())
+            {
+
+                if (tsdate.getRowCount() > 0)
```
bbimber: Is this really the best pattern here? Could you call TableSelector.getArrayList() (or whatever that method is called)? At absolute minimum shouldnt you call tsdate.exists(), which is better than getRowCount(), prior to requesting a Results? And I would check out TableSelector.forEach)() as an alternative to this.

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -1215,30 +1237,106 @@ private String checkOtherDrawsQuantity(String id, Date date, String rowObjectId,
...
-        //always report the most severe overage
-        if (!overages.isEmpty())
+        if (!overages.isEmpty() || !closeToThreshold.isEmpty())
+        {
+            StringBuilder errorMsgBuilder = new StringBuilder();
+
+            //always report the most severe overage
+            if (!overages.isEmpty())
+            {
+                errorMsgBuilder.append("Blood volume of ");
```
bbimber: I appreciate the benefit of StringBuilder over concatenation, but if you're going to do this, i think you can chain them, like:
```
errorMsgBuilder.append(rowQuantity)
                .append(" (")
                .append(overages.descendingSet().iterator().next())
                .append(" over ");
```
or something to that effect.

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -1215,30 +1237,106 @@ private String checkOtherDrawsQuantity(String id, Date date, String rowObjectId,
...
+            return errorMsgBuilder.toString();
+        }
+        else
         {
-            return "Blood volume of " + rowQuantity + " (" + overages.descendingSet().iterator().next() + " over " + interval + " days) exceeds the allowable volume of " + maxAllowable + " mL (weight: " + weight + " kg)";
+            return null; // No errors
         }
+    }
+
+    public double getThresholdLimit()
```
bbimber: I am guessing maybe this is how you intend to let WNPRC-specific code set this value? If so, I suggest making it more descriptive and adding javadoc to make the purpose of this method clearer. Same for any other public methods you add to this class

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -1215,30 +1237,106 @@ private String checkOtherDrawsQuantity(String id, Date date, String rowObjectId,
...
     {
         private String _objectId;
         private Date _date;
         private double _quantity;
+        private boolean _inTransaction;
 
         public BloodInfo() {}
 
-        public BloodInfo(String objectId, Date date, Double quantity)
+        public BloodInfo(String objectId, Date date, Double quantity, boolean isInTransaction)
```
bbimber, line 1350: Same question as above. It is necessary to set this?

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -1215,30 +1237,106 @@ private String checkOtherDrawsQuantity(String id, Date date, String rowObjectId,
...
+                errorMsgBuilder.append(interval);
+                errorMsgBuilder.append(" days) exceeds the allowable volume of ");
+                errorMsgBuilder.append(maxAllowable);
+                errorMsgBuilder.append(" mL (weight: ");
+                errorMsgBuilder.append(weight);
+                errorMsgBuilder.append(" kg).\n");
+            }
+
+            //report all the ones that are close to limit
+            for (double closeValue : closeToThreshold.descendingSet())
```
bbimber: Would you ever want to report multiple messages? Should you report the most severe, like above? Or try to distill multiple closeToThreshold values into one string message?

Also: should we differentiate between actual overages (which seems clearly like an error), and close-to-overage (which seems like maybe it should be a warning? While we just pass a basic string to the client, maybe we could modify this to prefix with "WARNING:", and have the JS side test for this, and convert this to a warning not error?

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
csebranek: you're right I was under the impression that we can't override the init, let me take a look at adding a script option instead.

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
csebranek: it is not true if it's strictly coming from the database, i didn't want to report the warning if it was a record that was already in the database, it seemed like those were still coming through

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
csebranek: I can add a javadoc

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
csebranek: I believe the users would like all of the messages, what do you mean by distill down?

For the error levels, we are doing something like that, but about one level less, meaning the close-to-overages are INFO level and the over the limits are WARN. We are controlling this error severity level on the JS side like you mention, [like this](https://github.com/LabKey/wnprc-modules/blob/893498276fb0436accbf5486b1353c892135cbcf/WNPRC_EHR/resources/scripts/wnprc_ehr/wnprc_triggers.js#L809) but by using the prefix "Limit" (instead of using a prefix like WARNING or ERROR, to not confuse later if those error severity levels are set differently for each center), and then using the prefix to set the error the threshold to INFO (only if there's no overages), while keeping the ones that are overages as WARNING. This lets me use the same string without refactoring a bunch of things. Blood draw requests never let us submit when there's a WARNING, unless it's a privileged user, afaik.

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
bbimber: To add to this: I would check whether just using a TableSelector.forEach() is sufficient. I am not sure there is any benefit to calling something like rowCount() or exists() first, followed by the query, since that is two separate SQL queries. It might be worth a small check on performance, but in general I would guess one query is better than two.

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
bbimber: Do you have a pointer to any code that is actually using this where inTransaction=FALSE?

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
bbimber: One more comment: when you add javadoc to the setThreshold() method (and I might rename that to make it more descriptive), can you please mention that true overages and near-overages will be returned as one array, but the consumer than separate them based on the message (or something to that effect). We should at least put some effort into making this halfway reasonable for a future developer to decipher how to use it. Maybe even put a link to your WNPRC github code in the javdoc?

To the other question about distilling errors. At least for actual overages, it is possible for a given blood record to cause multiple problems with different volumes, assuming there are multiple future draws in question. That existing code collapses what would be multiple redundant errors into a single message and reports the max volume of the overage. 

Unless I"m missing something, I am questioning whether this same pattern makes sense for near-overages. If a given blood record would put the limit within 2mL based on some record and also 1mL based on some other combination of records, is there a reason the user needs to see two different messages that only differ by their volume?

bbimber:
One additional comment: your changes to the TSVs mention tests, which is great. Do you have tests associated with this?

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
csebranek: yeah, here: https://github.com/LabKey/ehrModules/blob/18e14d2ebf8a2d05d8586b6dc9f146c20f579791/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java#L1201

bbimber: Alright - I think I get what you're doing. This is another case where I think it's important to leave documentation within the code on the purpose of these toggles. Can you add something around this, maybe in the BloodInfo class? This doesnt need to be formal javadoc, but some explanatory comments are a good idea

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
bbimber: Thanks. Documention is probably most important for getThresholdCheck(), and I would really rename the latter to something much more explanatory. TriggerHelper isnt blood-specific, and getThresholdCheck doesnt say anything about being blood-related. I would even consider doWarnForBloodNearOverages() or something very clear.

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -1278,6 +1376,16 @@ public void setQuantity(double quantity)
             _quantity = quantity;
         }
 
+        public void setIsInTransaction(boolean inTransaction)
```
bbimber: I understand the new variable, but do you actually need to expose a setter?

csebranek: probably not, this can be removed

csebranek:
> One additional comment: your changes to the TSVs mention tests, which is great. Do you have tests associated with this?

We have some tests on our end, I didn't go through adding them to the EHR but it could be done - the tests are again specific to WNPRC's error thresholds. I think I grabbed them from ONPRC's tests, or maybe it was a rough copy of the assignment tests there: https://github.com/LabKey/wnprc-modules/blob/893498276fb0436accbf5486b1353c892135cbcf/WNPRC_EHR/test/src/org/labkey/test/tests/wnprc_ehr/WNPRC_EHRTest.java#L3543 . I did enough to trigger the error messages come back, didn't bother to put them in an additional extra context (which probably should be refactored in EHR at some point, because putting everything through extraContext seems a bit legacy), they test the minimum.

bbimber:
> > One additional comment: your changes to the TSVs mention tests, which is great. Do you have tests associated with this?
> 
> We have some tests on our end, I didn't go through adding them to the EHR but it could be done - the tests are again specific to WNPRC's error thresholds. I think I grabbed them from ONPRC's tests, or maybe it was a rough copy of the assignment tests there: https://github.com/LabKey/wnprc-modules/blob/893498276fb0436accbf5486b1353c892135cbcf/WNPRC_EHR/test/src/org/labkey/test/tests/wnprc_ehr/WNPRC_EHRTest.java#L3543 . I did enough to trigger the error messages come back, didn't bother to put them in an additional extra context (which probably should be refactored in EHR at some point, because putting everything through extraContext seems a bit legacy), they test the minimum.

OK. copying the ONPRC API test is what I was going to suggest. What do you mean by "didn't bother to put them in an additional extra context"

Separate question since you brought it up: is there discussion about changes around how EHR uses the LabKey API? I completely agree that bolting on lots of information through extraContext is not ideal, but is there a specific solution proposed? What it really needs is something more like an EHR-specific API format, which probably needs an EHR-specific Select/Save API that extends the core APIs and handles EHR-specific information. There are other options as well, but this would be the shortest change form today to preserve a lot of that logic.

csebranek:
> > > One additional comment: your changes to the TSVs mention tests, which is great. Do you have tests associated with this?
> > 
> > 
> > We have some tests on our end, I didn't go through adding them to the EHR but it could be done - the tests are again specific to WNPRC's error thresholds. I think I grabbed them from ONPRC's tests, or maybe it was a rough copy of the assignment tests there: https://github.com/LabKey/wnprc-modules/blob/893498276fb0436accbf5486b1353c892135cbcf/WNPRC_EHR/test/src/org/labkey/test/tests/wnprc_ehr/WNPRC_EHRTest.java#L3543 . I did enough to trigger the error messages come back, didn't bother to put them in an additional extra context (which probably should be refactored in EHR at some point, because putting everything through extraContext seems a bit legacy), they test the minimum.
> 
> OK. copying the ONPRC API test is what I was going to suggest. What do you mean by "didn't bother to put them in an additional extra context"
> 
> Separate question since you brought it up: is there discussion about changes around how EHR uses the LabKey API? I completely agree that bolting on lots of information through extraContext is not ideal, but is there a specific solution proposed? What it really needs is something more like an EHR-specific API format, which probably needs an EHR-specific Select/Save API that extends the core APIs and handles EHR-specific information. There are other options as well, but this would be the shortest change form today to preserve a lot of that logic.

I was looking at the [assignment tests](https://github.com/LabKey/onprcEHRModules/blob/develop/onprc_ehr/test/src/org/labkey/test/tests/onprc_ehr/ONPRC_EHRTest.java#L332C37-L332C61) in ONRPC and was a bit confused why there were records being placed in the extracontext for testing validation - I was making the false assumption that maybe there was a better way to test the validation for multiple records here since all of the blood records are kept in an extra context, but I honestly do not know, I think now looking back that the assignments is expecting something to fail there, so no worries.

For your other question, no proposed solution here other than not stuffing everything into extraContext. We're starting to develop some newer EHR frontend, we were directed by LabKey to use the older APIs which support extraContext, I do not know if there will be a EHR-specific API format that might be a question for @labkey-martyp. Currently it looks like the new JavaScript APIs have extraContext as [experimental](https://labkey.github.io/labkey-api-js/interfaces/Query.SaveRowsOptions.html#extraContext). I agree that an EHR specific saveRows would be the cheapest option here.

bbimber:
@csebranek: If memory serves, the point of bloodInTransaction, assignmentInTransction, etc., is for the situation where you have an open form with lots of records, including many that do not exist in the DB.  The client is constantly sending individual records to the server for validation, one record at a time. That is important for speed; however, it has the disadvantage of the server not always bring aware of all the records about to be saved. When the server needs more context, the ExtJS store code packages a map of all client side records (like blood, weight, or assignments) and posts that to the server along with the row in question. This gives the server enough information to merge server-side records and all incoming client-side records, which lets the server-side validation accurately predict things like blood volumes. If this didnt happen, the client-side warnings would be wrong, and they would only trigger once the user tried to submit the whole form. 

Anyway, long explanation, but the goal should be for your java API-only test case to create JSON that adequately mirrors whatever JSON is going to be created by the ExtJS forms. If all you need is to provide the array of rows, great. Since the other pieces are probably tested elsewhere, i think a simple test is adequate (i.e. not adding bloodInTransaction)

bbimber:
Regarding the API question: in my opinion, that's a really fundamental and important question that ought to at least be discussed well. Some key properties of EHR's data entry, like the real-time feedback, depend on features that are admittedly bolted over LabKey's query APIs. I could see multiple paths that could be argued:

- Keep the status quo, but write better client-side typescript classes and maybe some helpers. The would perform the functions currently handled in the EHR/ExtJS store layer. These would at least enforce some structure and document the expected JSON, even if this is still just poked into extraContext and passed over LabKey's SaveRows API. This typescript layer could both create and parse the API JSON, giving more structure and sanity to it. It might be a middle ground between entirely reworking the server-side API. This basically sandboxes all the LabKey/EHR-specific traffic and exposes a clean layer upon which people could build their new UI.

- There is an argument to replace the server-side API endpoint with an EHR-specific one, with an EHR-specific JSON format. This has the advantage of structuring the server-side of the JSON better. This would also mean we could do a better job of typing extraContext() and being more deliberate and explicit on what properties are or are not supported. This might not want to blindly reimplement what's used now.

- Related to the above, if features like the immediate error feedback are going to continue to be used, re-thinking an EHR-specific API that is more appropriate specifically to this task might be worth discussion. 

just my two cents here...

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
csebranek: @bbimber any idea how to propagate these script options through to the java helper?

csebranek: Actually I think setCenterCustomProps might do the trick, let me see

csebranek: @bbimber I found something called setCenterCustomProps which might do the trick, it doesn't look like setScriptOptions gets propagated over

bbimber: If I understand you question I would think about it like this. There's a tradeoff:

- The way you implemented this now is to expose a setter on the java trigger helper class for the 'reportNearOverage' property and a setter to set the threshold value. For a center to opt-in to this behavior, they need to do that you did and override the core blood.js onUpsert() code with their own logic. 

- Option 2 would be to be more like how some existing support properties work, like 'allowDeadIds', 'doStandardProtocolCountValidation', etc. It would not be crazy to support properties related to the blood-near-overage behavior. If we did that, the core blood.js code would read that property, call the TriggerScriptHelper setters as appropriate, and parse the resulting messages into error/info as appropriate. Then the WNPRC module (or anyone else wanting this), would just need to register code vaguely like this:
```
EHR.Server.TriggerManager.registerHandlerForQuery(EHR.Server.TriggerManager.Events.INIT, 'study', 'blood', function(event, helper, EHR) {
        helper.setScriptOptions({
            reportBloodDrawNearOverage: true,
            bloodDrawNearOverageThreshold: 4
        });
    });
```

And then core EHR code would deal with the rest. That's probably how I would do it. The goal would be to keep as much of the codepath as reasonable shared between centers. 

That said, as long as this is well documented and explained in EHR's code, I think the split you have between EHR and WNPRC is reasonable as-is.

csebranek: it looks like setScriptOptions does not get propagated to the JS layer, but there is a setCenterCustomProps that does, so I went with trying that out.

it seems like when you do something like:

```
EHR.Server.TriggerManager.registerHandlerForQuery(EHR.Server.TriggerManager.Events.INIT, 'study', 'blood', function (helper, scriptErrors, row, oldRow) {
            helper.setCenterCustomProps({
                doWarnForBloodNearOverages: true,
                bloodNearOverageThreshold: 4.0 
            })
        });
```

introducing the JS layer for this seems to work ok, but you then encounter weird casting issues.
the when I send 4.0 as the val of bloodNearOverageThreshold, the JS interpreter thinks bloodNearOverageThreshold is an integer, so it needs to be converted back to a double for the calculation. but for example if I send a 4.5, it will get interpreted as a double.

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -1860,7 +1860,7 @@ else if (lastDeparture != null)
                 }
                 else if (hasCustomDepartureStatus)
                 {
-                    status = _centerCustomProps.get("departureStatus");
+                    status = _centerCustomProps.get("departureStatus").toString();
```
bbimber: would this ever NPE, such as if departureStatus wasnt present? Do you need the toString(), and what about the null-safe String.valueOf() instead?

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
csebranek: I replaced this with a forEach()

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
csebranek: I chained them now

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
csebranek: Added a javadoc and a new name for getting the limit: getBloodNearingOveragesThreshold() based on setting custom props, what do you think about using _centerCustomProps?

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
csebranek: Yes, added a comment why the isInTransaction is needed

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
csebranek: I added a comment to the new class member

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
csebranek: I think hasCustomDepartureStatus checks for an NPE

csebranek: I guess technically _centerCustomProps could still have the key departureStatus and still be null, using the null safe option instead

bbimber: it's containsKey(), not a null check. it's admittedly unlikely to have a property that has been set to null, but that case would NPE here.

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
bbimber: I dont mean to introduce a bunch of extra hurdles here, and I could see multiple ways of approaching this being fine. 

First: the reason you have having problems with setCenterCustomProps() is that this is getting passed from the JS layer to the java layer as a Map<String, String>. You could probably do a null-check and then call Double.parseDouble(String.valueOf(<RAW_VALUE>)) or something similar, with error catching for whatever ParseException that method throws in case a non-numeric value is provided. The JS code in ScriptHelper that's passing centerCustomProps from JS to java isnt doing any handling of data type, and i think this could give you weird behaviors. I would suggest string casting and re-parsing like I suggest above if you use this route.

Second, I would not expect all of the values from the standard "setScriptOptions()" to automatically be passed to the java trigger layer. If you went with that route, you would need to additionally add a line in core EHR's blood.js onUpsert() code to call something to the effect of: 
```
getJavaHelper().doWarnForBloodNearOverages(triggerHelper.doWarnForBloodNearOverages())
getJavaHelper().setBloodNearOverageThreshold(triggerHelper.getBloodNearOverageThreshold())
```
In other words, the scriptProperties are only copied from the JS into the java layer if some code actually does it. Most of them are never used in the java code, or are only copied in when actually needed. The primary reason the scriptProperties object exists is to get at least some documentation and minimal discoverability around what config options are supported without needing to read into the depths of the code. 

Anyway, what you're doing with setCenterCustomProps is at least as valid as some of the other things going through it.

csebranek: where do you see it being passed as Map<String, String>? I changed it from Map<String,String> to Map<String, Object> _centerCustomProps; and those doubles from the JS side are still getting sent over as Integers, so I did something like:
```
Object theVal = _centerCustomProps.get("bloodNearOverageThreshold");
if (theVal instanceof Integer)
{
  Integer theValInt = (Integer) theVal;
  return theValInt.doubleValue();
}
```

bbimber: OK, apologies. I must have been mixing up what I looked at locally (which doesnt have your changes) with this. 

That all being said, my experience with dealing with JS objects being ported into java is that it's really easy to get obscure bugs, I so I generally avoid it. For example, you might have issues with Rhino casting the JS integer/double into Number or some different java classes. I am not running your code locally, so I am just saying this from experience, not testing your code. Anyway, if you think you need to do 'instanceof' tests, I strongly recommend doing what I suggested above and using Double.parseDouble(String.valueOf(<theValue>). It's not really any less awkward than instanceOf testing, and should be robust to however rhino decides to wrap the JS data.

bbimber: Also, while your example is 4.0, and thus double->int->double doesnt matter, would your Integer casting actually work if the value wasnt an integer?

csebranek: no worries I appreciate the feedback. So As I understand String.valueOf() expects a string, but if my Map<String, Object> values are "objects" it looks like Rhino or whatever is doing the translation sometimes works, in that if I pass in a boolean, it gets casted in as a boolean for the most part, but passing in a double of 4.0 it gets casted to an Integer then I have to cast it back to an int, otherwise just use the default value.

so the full codeblock would be something like:

```
 public double getBloodNearingOveragesThreshold()
    {
        Object theVal = _centerCustomProps.get("bloodNearOverageThreshold");
        if (null != theVal)
        {
            if (theVal instanceof Integer)
            {
                Integer theValInt = (Integer) theVal;
                return theValInt.doubleValue();
            }
            else if (theVal instanceof Double)
            {
                return (Double) theVal;
            }
        }
        return _bloodNearingOveragesThresholdDefaultValue;
    }
```


I am not sure how much data checking should be done here, but maybe I can just add it to the documentation to say this prop's value is expected to be a double?

bbimber: To break down those points:
- My impressions of rhino could have bene fixed in newer versions, but in the past I found it to be inconsistent as far as how it translates JS objects into java, and quirky variations in your JS objects would change the result on the java side. Therefore I generally trust it as little as possible, and untyped Map<String, Object>s or similar should be avoided here, in my opinion.

- You wrote: "so As I understand String.valueOf() expects a string" - no, it does not and that's exactly why I propose it here. The reason I'm suggesting that pattern is that I think it's the most robust serialization that will be tolerant to whatever quirky variations happen in your data. 

Anyway, what you have is fine - it's not worth going back and forth. I would suggest adding an error check:
```
            else if (theVal instanceof Double)
            {
                return (Double) theVal;
            }
else {
    // Either throw or log an error here
}

```

csebranek: yeah, sorry i meant the variant of Double.valueOf() expects strings. The reason I chose the untyped Map<String,Object> was because it seemed odd to go from double -> string -> double. 

But anyway I have an even bigger problem with this setScriptOptions codepath (in this case setCenterCustomProps). Since EHR's blood.js does not have a registered INIT trigger, we cannot override it with one. Therefore we need to override it only with an onInit which means duplicating all of the code in the current onInit just to set some custom options for WNPRC. Can you please clarify this behavior @labkey-martyp and let us know what is the best approach for setting a custom "script option" for this usecase? Before going this route I just had exposed a public setter method in the trigger script helper that seemed the fastest way, but setting these script options is now causing me headaches.

bbimber: why do you need to override it? can't you just register you handler (which should be called after core EHR's onInit), and set whatever you need? why would you need (or want) to replace the core's onInit?

csebranek: my fault, I was registering the INIT handler incorrectly, but I was not getting any error messages so I was confused about what was going wrong. It looks to be working now with those custom script options. logging an error message now too if a null val is found.

Any other requested changes for this PR?

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -1215,30 +1223,128 @@ private String checkOtherDrawsQuantity(String id, Date date, String rowObjectId,
...
+                               .append(maxAllowable)
+                               .append(" mL (weight: ")
+                               .append(weight)
+                               .append(" kg).\n");
+            }
+
+            //report all the ones that are close to limit
+            for (double closeValue : closeToThreshold.descendingSet())
+            {
+                errorMsgBuilder.append("Limit notice! Blood volume of ")
```
bbimber, line 1261: @csebranek: i just noticed this comment was closed. I question whether there is any point in looping over the list of near-overages. It is possible for a given blood draw to have multiple overages and/or multiple near-overages, since it could overlap with varying numbers of future records. With overages, the code just reports the max(overage), rather than giving one message for every overage.

Is there any reason you shouldnt do the exact same thing here? In what instance would having multiple nearly identical messages be preferable to reporting the max(closeValue)?

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -1215,30 +1223,128 @@ private String checkOtherDrawsQuantity(String id, Date date, String rowObjectId,
...
+        {
+            if (theVal instanceof Integer)
+            {
+                Integer theValInt = (Integer) theVal;
+                return theValInt.doubleValue();
+            }
+            else if (theVal instanceof Double)
+            {
+                return (Double) theVal;
+            }
```
bbimber, line 1313: can you please either throw or log an error message if theVal is neither an integer or double? otherwise it will just get silently ignored.

csebranek: added a log message since it just defaults to a default value and can continue to run

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
csebranek: I see your point, not much use in having error messages representing nearly the same information, i distilled it down to one

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
bbimber: Thanks. Two things:

1) I think if the module supplies a value, but that value is invalid, it should be error(), not warn()
2) The second warn() right below that doesnt seem right to me. If we except the developer to provide a threshold, we should just enforce that and not default to anything. If we think it's valid for the develop to just accept the default, then I dont think we should warn() about it - that's just noise. Maybe there is an argument for debug() logging.

csebranek: maybe then both cases should just throw an error if a bad or empty value is found for this, added another commit

bbimber:
@labkey-martyp  or @labkey-jeckels: is there a way to make this run against the various center-specific tests, or did this perhaps already happen and maybe it's not being reported here? I dont have complete access to TeamCity anymore so I might be missing something that's obvious from that side.

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -1215,30 +1223,139 @@ private String checkOtherDrawsQuantity(String id, Date date, String rowObjectId,
...
```
labkey-martyp:
```suggestion
            if (theVal instanceof Integer theValInt)
            {
                return theValInt.doubleValue();
            }
            else if (theVal instanceof Double theValDouble)
            {
                return theValDouble;
            }
```

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -134,10 +134,12 @@ public class TriggerScriptHelper
      *  <li>departureStatus - center specific custom status for animal departures</li>
      */
     @NotNull
-    private static final Map<String, String> _centerCustomProps = new HashMap<>();
+    private static final Map<String, Object> _centerCustomProps = new HashMap<>();
 
     private static final Logger _log = LogHelper.getLogger(TriggerScriptHelper.class, "Server-side validation of EHR data insert/update/deletes");
 
+    private static double _bloodNearingOveragesThresholdDefaultValue = 4.0; //in mL
```
labkey-martyp: Is this being used somewhere? I'm not finding any references to it.

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -1215,30 +1223,139 @@ private String checkOtherDrawsQuantity(String id, Date date, String rowObjectId,
...
+                return (Double) theVal;
+            }
+            else
+            {
+                throw new RuntimeException("TriggerScriptHelper.getBloodNearingOveragesThreshold invalid value found for _centerCustomProps.bloodNearOverageThreshold.");
+            }
+        }
+        else
+        {
+            throw new RuntimeException("TriggerScriptHelper.getBloodNearingOveragesThreshold no value found for _centerCustomProps.bloodNearOverageThreshold, please supply a value.");
```
labkey-martyp:
```suggestion
            throw new RuntimeException("TriggerScriptHelper.getBloodNearingOveragesThreshold no value found for _centerCustomProps.bloodNearOverageThreshold. If doWarnForBloodNearOverages is set to true, then bloodNearOverageThreshold must also be set.");
```

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -1215,30 +1223,139 @@ private String checkOtherDrawsQuantity(String id, Date date, String rowObjectId,
...
+                Integer theValInt = (Integer) theVal;
+                return theValInt.doubleValue();
+            }
+            else if (theVal instanceof Double)
+            {
+                return (Double) theVal;
+            }
+            else
+            {
+                throw new RuntimeException("TriggerScriptHelper.getBloodNearingOveragesThreshold invalid value found for _centerCustomProps.bloodNearOverageThreshold.");
```
labkey-martyp:
```suggestion
                throw new RuntimeException("TriggerScriptHelper.getBloodNearingOveragesThreshold invalid value found for _centerCustomProps.bloodNearOverageThreshold. Required type is a double.");
```

labkey-martyp (approved):
This looks good to me @csebranek, just a few small suggestions.

@bbimber this is not causing any other center specific tests to fail.

labkey-jeckels:
@bbimber do you see the two Daily EHR suites (Postgres and SQLServer) reported here in GitHub? I do, but I'm not sure which permissions control that. They include the center-specific EHR tests based on the DB in use.

bbimber (approved):

bbimber:
OK, got it. @csebranek, thanks for the changes. from my end this seems fine.

ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
csebranek: It used to be, but since we're throwing errors instead of relying on a default val it is no longer being used. I removed this

PR #775, opened 2024-05-01
Optional async demographic provider updates

labkey-martyp:
#### Rationale
Provide a way for individual demographic providers to be run asynchronously in another thread. Doing this shortens data insertion times and avoids trigger timeout limits in extreme cases.

#### Related Pull Requests
* https://github.com/LabKey/onprcEHRModules/pull/996

#### Changes
* Add isAsync option to DemographicProvider
* If isAsync is true, split provider cache update to a separate postcommit task job

### DIFF
--- a/ehr/src/org/labkey/ehr/demographics/EHRDemographicsServiceImpl.java
+++ b/ehr/src/org/labkey/ehr/demographics/EHRDemographicsServiceImpl.java
@@ -25,6 +25,7 @@ import org.labkey.api.cache.CacheManager;
 import org.labkey.api.data.CompareType;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerManager;
+import org.labkey.api.data.DbScope;
 import org.labkey.api.data.SimpleFilter;
 import org.labkey.api.data.TableInfo;
 import org.labkey.api.data.TableSelector;
@@ -72,7 +73,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
-import java.util.concurrent.TimeUnit;
 
 /**
  * User: bimber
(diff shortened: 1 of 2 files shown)


ankurjuneja (approved):

ehr/src/org/labkey/ehr/demographics/EHRDemographicsServiceImpl.java
```diff
@@ -339,7 +340,16 @@ private void doUpdateRecords(Container c, User u, List<Pair<String, String>> cha
 
             for (DemographicsProvider p : needsUpdate)
             {
-                updateForProvider(defaultSchema, p, ids, true, async);
+                // If not already in an async thread, and the provider is async, defer just this provider update to an async thread
+                if (!async && p.isAsync())
+                {
+                    JobRunner.getDefault().execute(TimeUnit.SECONDS.toMillis(30), () -> {
```
labkey-jeckels: Why the 30 second delay?

Should we switch all updates to be async?

Also, you'll need to set the query environment in the async running. See the setup in the async code in reportDataChange()

labkey-jeckels (changes requested):

ehr/src/org/labkey/ehr/demographics/EHRDemographicsServiceImpl.java
bbimber: I'm confused by this proposal in combination with this older change: https://github.com/Labkey/ehrModules/commit/f4916a1016bf37d0c80d4ecadce0c2969bfb5725. This existing argument called 'async' just seems to control where one line is logged or not? Why do we have/need two different 'async' properties and should that be more unified?

I can imagine there are good reasons to push at least some of this work async. I would be concerned about flipping 100% of this without a reasonable amount of testing to ensure there are not existing codepaths that make the assumption that the cache is accurate within the transaction. 

Finally, since this is a property of the DemographicsProvider, wouldnt it make more sense to contain this logic within updateForProvider(), rather than doUpdateRecords()?

labkey-martyp: Ah ok, I missed the environment. I wasn't aware that was necessary. Added.

30 second delay just to ensure the transaction that kicked off the cache update is complete, including any updates to the demographic providers. Seemed like a fairly short wait to ensure there are no race conditions.

I think switching all updates to be async would be a big improvement for the EHR, however that is a little broader stroke than I had hoped for here as it would require testing all demographic providers and possibly making updates like I had to do in the related PR.

labkey-martyp: @bbimber The async parameter in the commit you referenced is only used for ETLs and the centers who are using it have the logic in their providers to handle the async operation (basically the state of the demographic providers in transaction vs out of transaction).

updateForProvider is called recursively so I don't think we want to execute the job in there.

labkey-jeckels (approved):

PR #778, opened 2024-05-07
Extensible lab_tests table

labkey-martyp:
#### Rationale
Make lab_tests and the virtual tables evolved from this table (chemistry_tests, hematology_tests, etc.) extensible.

#### Changes
* Add LSID column and LSID fk
* Override getDomainURI to get correct domain for virtual tables


labkey-jeckels (approved):

labkey-adam:
Binal recently back-ported the EHR schema version bumps to 23.11, to accommodate potential EHR (and other) scripts coming from that release branch. This PR effectively closes the door on all future EHR scripts from that branch. That's probably fine, but we'll need to remember.

PR #781, opened 2024-05-07
Move demographic recache to post commit task

labkey-martyp:
#### Rationale
The current demographic provider recache after creating a new demographic record can get into race condition with long running trigger scripts. Update this recache to be a post commit task.

Move some test helpers to EHR.

#### Changes
* Make demographic recache a post commit task
* Remove delay from job as recache is now running after the transaction commits
* Add some test helpers


labkey-jeckels (approved):
Please check the test failures to see if any are related.

PR #791, opened 2024-06-18
Improvements to the protocol validation for animal assignments

csebranek:
#### Rationale
Currently the system does not check for a couple cases when assigning animals to a project/protocol. This validation is called from EHR's assignment.js trigger script. One missing case is if the species is not at all listed in the protocol_counts table for the given protocol (in this case it is assumed to be zero animals allowed). And another case is if there's a max number of animals that are Macaques, since that is the genus for Cynomolgus and Rhesus and also can have a cap e.g., if a protocol allows only 10 Macaque meaning it could be 5 Cyno and 5 Rhesus or another mix but no greater than 10 total. Also the protocolTotalAnimalsBySpecies.sql should be checking against the protocol_counts table for the most part, since that's where the species level counts are located, at least for WNPRC, I am assuming this is the case for other centers.

#### Related Pull Requests
* 

#### Changes
* Join with the protocol_counts table
* Add checks for Macaque 
* Add check for when a species is not listed at all on a protocol

### DIFF
--- a/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
+++ b/ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
@@ -2256,11 +2256,20 @@ public class TriggerScriptHelper
 
         //find the total animals previously used by this protocols/species
         TableInfo ti = QueryService.get().getUserSchema(getUser(), getContainer(), "ehr").getTable("protocolTotalAnimalsBySpecies");
-        SimpleFilter filter = new SimpleFilter(FieldKey.fromString("species"), PageFlowUtil.set(ar.getSpecies(), "All Species"), CompareType.IN);
+        String animalSpecies = ar.getSpecies();
+        SimpleFilter filter;
+        if (("Rhesus".equals(animalSpecies) || "Cynomolgus".equals(animalSpecies)))
+        {
+             filter = new SimpleFilter(FieldKey.fromString("species"), PageFlowUtil.set(animalSpecies, "All Species", "Macaque"), CompareType.IN);
+         } else
+         {
+             filter = new SimpleFilter(FieldKey.fromString("species"), PageFlowUtil.set(animalSpecies, "All Species"), CompareType.IN);
+         }
         filter.addCondition(FieldKey.fromString("protocol"), protocol);
         TableSelector ts = new TableSelector(ti, filter, null);
         final List<String> errors = new ArrayList<>();
         final String ALL_SPECIES = "All Species";
+        final boolean[] noSpeciesListedOnProtocol = {true};
         ts.forEach(new Selector.ForEachBlock<ResultSet>()
         {
             @Override
@@ -2294,8 +2303,8 @@ public class TriggerScriptHelper
                         {
                             continue;
                         }
-
-                        if (!ALL_SPECIES.equals(species))
+                        //we don't want to exit the animal count if ar.species() is cyno or rhesus and the protocol's species value is macaque.
+                        if (!ALL_SPECIES.equals(species) && !("Rhesus".equals(ar.getSpecies()) || "Cynomolgus".equals(ar.getSpecies()) && "Macaque".equals(species)))
                         {
                             //find species
                             AnimalRecord ar = getDemographicRecord(id);
@@ -2306,8 +2315,16 @@ public class TriggerScriptHelper
                         }
 
                         animals.add(id);
+
+                        AnimalRecord ar = getDemographicRecord(id);
+                        if (ar.getSpecies().equals(species))
+                        {
+                            noSpeciesListedOnProtocol[0] = false;
+                        }
                     }
                 }
+
+
 
                 int remaining = totalAllowed - animals.size();
                 if (remaining < 0)
@@ -2316,6 +2333,11 @@ public class TriggerScriptHelper
                 }
             }
         });
+
+        if (noSpeciesListedOnProtocol[0])
+        {
+            errors.add("Species not allowed on protocol: " + protocol);
+        }
 
         return StringUtils.join(errors, "<>");
     }
--- a/ehr/resources/queries/ehr/protocolTotalAnimalsBySpecies.sql
+++ b/ehr/resources/queries/ehr/protocolTotalAnimalsBySpecies.sql
@@ -30,6 +30,10 @@ FROM
 
         UNION
 
+        /*
+         * Calculate the macaque totals as well, as the sum of Cynos and Rhesus
+         */
+
         SELECT
             coalesce(p.protocol, pa.protocol) as protocol,
             'Macaque' as species,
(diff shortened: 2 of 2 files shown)


ehr/src/org/labkey/ehr/utils/TriggerScriptHelper.java
```diff
@@ -2256,11 +2256,20 @@ public String verifyProtocolCounts(final String id, Integer project, final List<
...
+             filter = new SimpleFilter(FieldKey.fromString("species"), PageFlowUtil.set(animalSpecies, "All Species", "Macaque"), CompareType.IN);
+         } else
+         {
+             filter = new SimpleFilter(FieldKey.fromString("species"), PageFlowUtil.set(animalSpecies, "All Species"), CompareType.IN);
+         }
         filter.addCondition(FieldKey.fromString("protocol"), protocol);
         TableSelector ts = new TableSelector(ti, filter, null);
         final List<String> errors = new ArrayList<>();
         final String ALL_SPECIES = "All Species";
+        final boolean[] noSpeciesListedOnProtocol = {true};
```
bbimber, line 2272: This is a minor comment and there's nothing wrong with using an array, but I would say AtomicBoolean is a little clearer. In my opinion is makes it more obvious what the purpose of using that object it, relative to making an array purely for the purpose of using a java quirk.

bbimber (approved):
So long as all the test pass, I dont really have much to say one way or the other here. 

A lot of this logic is probably purely WNPRC; however, due to the fact that WNPRC was the first center the core EHR contains a lot of WNPRC-specific business logic anyway. If we were starting from scratch it would probably be better to separate more of this into the center modules, but that's clearly way out of scope here and not worth it. 

I had one minor comment around java's quirky use of single-element arrays, which I dont feel that strongly about if others disagree.

labkey-martyp (changes requested):
Hi Chad, this is pretty specific to WNPRC and has some hard coded values for species that won't match most PRCs. In our meeting today lets discuss moving this logic into the WNPRC module.

bbimber:
@csebranek: OK, if this is getting more review, here are some more thought-through comments:

1) On protocolTotalAnimalsBySpecies.sql:
- You probably know that you can add a file called protocolTotalAnimalsBySpecies.sql into the WNPRC EHR module and this will override the one in EHR? This seems like an easy change you can do in your module to avoid changing business logic in EHR itself.
- You added a sub-query that looks up approval date from ehr.protocol rather than re-add this into the query. Presumably that would be needed if you have an instance where an animal was assigned to study.assignments, but there is no record in protocol_counts. This is picky, but would it be better to change that sub-query to: "coalesce(p.approve, pa.protocol.approve) as approve"? you'd then also need to change the group_by to: "GROUP BY coalesce(p.protocol, pa.protocol, pa.protocol.approve), p.approve". Below is SQL with my suggestions on this. I dont have a local EHR instance to test this, but I think it would be valid SQL: 
[protocolTotalAnimalsBySpecies.sql.txt](https://github.com/user-attachments/files/15929158/protocolTotalAnimalsBySpecies.sql.txt)
- Despite what I wrote above about putting this into WNPRC_EHR, I'm pretty sure this one change is reasonable to add into the core EHR, since it does make 'approve' non-null for a condition that shouldnt ever happen, but i guess could happen
- That being said, my suggestion below might supersede any need to change this script
- If you do propose changes to core files, can you please try to keep whitespace changes minimal? it makes it much easier to view the diff without extraneous whitepsace edits that are not intentionally added.

2) There are a lot of flaws with the check being performed in the core EHR. In addition to what you point out, this check does not account for assignment date, and as far as I can tell it doesnt thoroughly check for existing DB records in assignments. ONPRC overrides all this logic, such as here: https://github.com/LabKey/onprcEHRModules/blob/272b2bc3418f2748a84578c373cb411adcb0f280/onprc_ehr/src/org/labkey/onprc_ehr/query/ONPRC_EHRTriggerHelper.java#L1569. Would it just make more sense for WNPRC to implement something parallel in the WNPRC_EHR module, patterned after what ONPRC does? I believe you could copy/adapt the method in ONPRC_EHR (which does contain some logic to hard-code species to roll-up)? If you clone our code you get the added benefit of correctly handling date of assignment. I think across center they would probably benefit from trying to implement a more complete check in core EHR, but as I noted earlier that would take some thought and not something I'm prepared to try to touch.

csebranek:
> @csebranek: OK, if this is getting more review, here are some more thought-through comments:
> 
> 1. On protocolTotalAnimalsBySpecies.sql:
> 
> * You probably know that you can add a file called protocolTotalAnimalsBySpecies.sql into the WNPRC EHR module and this will override the one in EHR? This seems like an easy change you can do in your module to avoid changing business logic in EHR itself.
> * You added a sub-query that looks up approval date from ehr.protocol rather than re-add this into the query. Presumably that would be needed if you have an instance where an animal was assigned to study.assignments, but there is no record in protocol_counts. This is picky, but would it be better to change that sub-query to: "coalesce(p.approve, pa.protocol.approve) as approve"? you'd then also need to change the group_by to: "GROUP BY coalesce(p.protocol, pa.protocol, pa.protocol.approve), p.approve". Below is SQL with my suggestions on this. I dont have a local EHR instance to test this, but I think it would be valid SQL:
>   [protocolTotalAnimalsBySpecies.sql.txt](https://github.com/user-attachments/files/15929158/protocolTotalAnimalsBySpecies.sql.txt)
> * Despite what I wrote above about putting this into WNPRC_EHR, I'm pretty sure this one change is reasonable to add into the core EHR, since it does make 'approve' non-null for a condition that shouldnt ever happen, but i guess could happen
> * That being said, my suggestion below might supersede any need to change this script
> * If you do propose changes to core files, can you please try to keep whitespace changes minimal? it makes it much easier to view the diff without extraneous whitepsace edits that are not intentionally added.
> 
> 2. There are a lot of flaws with the check being performed in the core EHR. In addition to what you point out, this check does not account for assignment date, and as far as I can tell it doesnt thoroughly check for existing DB records in assignments. ONPRC overrides all this logic, such as here: https://github.com/LabKey/onprcEHRModules/blob/272b2bc3418f2748a84578c373cb411adcb0f280/onprc_ehr/src/org/labkey/onprc_ehr/query/ONPRC_EHRTriggerHelper.java#L1569. Would it just make more sense for WNPRC to implement something parallel in the WNPRC_EHR module, patterned after what ONPRC does? I believe you could copy/adapt the method in ONPRC_EHR (which does contain some logic to hard-code species to roll-up)? If you clone our code you get the added benefit of correctly handling date of assignment. I think across center they would probably benefit from trying to implement a more complete check in core EHR, but as I noted earlier that would take some thought and not something I'm prepared to try to touch.

thanks for the suggestions to the SQL, we moved this code into the WNPRC repo, so I'm going to close this PR.

PR #795, opened 2024-06-21
Protocol counts verification

labkey-martyp:
#### Rationale
Move getProtocolForProject function from TriggerScriptHelper to EHRService for use by custom module triggers.

#### Related Pull Requests
* https://github.com/LabKey/wnprc-modules/pull/606

#### Changes
* Move getProtocolForProject and updateCachedProtocol to EHRService and update references to this function


labkey-jeckels (approved):
I recommend moving `updateCachedProtocol()` to `EHRService` as well so we don't have two different places mutating the same cache and having to agree on the cache keys, etc.

csebranek (approved):

PR #797, opened 2024-07-11
Add file repository in animal history reporting

ankurjuneja:
#### Rationale
This PR adds the file repository under animal history reporting. 

#### Related Pull Requests
* https://github.com/LabKey/nircEHRModules/pull/346


bbimber:
@labkey-adam: sorry, bad auto-complete when i first entered this comment. please disregard that notification. 

@ankurjuneja: does this PR really need to live in core EHR? is there any reason this cant be put into the nirc module? There's examples of other center modules doing analogous things.

labkey-bpatel:
> @labkey-adam: sorry, bad auto-complete when i first entered this comment. please disregard that notification.
> 
> @ankurjuneja: does this PR really need to live in core EHR? is there any reason this cant be put into the nirc module? There's examples of other center modules doing analogous things.

@bbimber Is there a reason it can't be in core EHR? There is a lot of duplicated code everywhere, so we are trying to minimize that whenever it is a little easy to do so, like in this case. We will have to do some minor refactoring (to configure default folders) for a couple other centers using this exact functionality.

bbimber:
@labkey-bpatel I agree this does not have a black and white answer and one could argue one direction or another. 

what about putting into a standalone JS file? does that mitigate both concerns? this would let centers that want it register a client dependency without adding more JS code that is loaded 100% of the time for everyone.

labkey-bpatel:
> @labkey-bpatel I agree this does not have a black and white answer and one could argue one direction or another.
> 
> what about putting into a standalone JS file? does that mitigate both concerns? this would let centers that want it register a client dependency without adding more JS code that is loaded 100% of the time for everyone.

Sure, that's a good idea, esp. for ease of maintainability for all the new implementations that is currently in progress. I am sure we will be adding more shared code there. Thanks for the suggestion!

ehr/resources/web/ehr/sharedReports.js
```diff
@@ -0,0 +1,170 @@
...
+                                            {
+                                                xtype: 'button',
+                                                style: 'margin-left: 10px;',
+                                                border: true,
+                                                text: 'Create Folders',
+                                                handler: function () {
+                                                    animalFolder.createDirectory({
+                                                        path: "/",
+                                                        success: function () {
+                                                            const folders = [
```
labkey-martyp, line 83: These need to be configurable. Can you set these as the default option but pass in a parameter that can override them. You'll need to wrap this in the NIRC module in order to pass in custom folders.

labkey-martyp: As discussed sticking with these for this initial implementation

labkey-martyp (approved):

PR #808, opened 2024-08-08
Add field to species table, isNHP. Default to true

ankurjuneja:
#### Rationale
This adds isNHP column to ehr.species table.

#### Related Pull Requests
* https://github.com/LabKey/johnsHopkinsEHRModules/pull/308

#### Changes
*

### DIFF
--- a/ehr/resources/schemas/dbscripts/postgresql/ehr_lookups-24.002-24.003.sql
+++ b/ehr/resources/schemas/dbscripts/postgresql/ehr_lookups-24.002-24.003.sql
@@ -1,0 +1,1 @@
+ALTER TABLE ehr_lookups.species ADD COLUMN isNHP BOOLEAN DEFAULT TRUE;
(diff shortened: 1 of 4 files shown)


bbimber:
Two questions,

First, would it be more generalizable to add an lsid column and just make this table extensible? I understand this one change could be argued as reasonable, but might this be a better precedent? 

Second, if you do go this route, should you add upgrade script code to retrofit existing data and change from null to true (I don't know how the databases handle adding new columns).

labkey-martyp (approved):

labkey-martyp:
> Two questions,
> 
> First, would it be more generalizable to add an lsid column and just make this table extensible? I understand this one change could be argued as reasonable, but might this be a better precedent?
> 
> Second, if you do go this route, should you add upgrade script code to retrofit existing data and change from null to true (I don't know how the databases handle adding new columns).

Yeah we discussed making this table extensible but this seemed like a reasonable change as groups are starting to support non-NHP species. I didn't test it but it should set all pre-existing rows to true. Anyone wanting to use the column will need to set it to false for non-NHP species.

PR #818, opened 2024-08-26
Track blood draw warning limits in the species table

csebranek:
#### Rationale
Right now the EHR trigger script will warn the user if they are approaching the blood limit for an animal but it's set globally for all animals in a trigger script. This PR will move this into the database so that the warning threshold can be set per species.

#### Related Pull Requests
* 

#### Changes
* Add blood threshold warning column to ehr_lookups.species
* Select from above column for the warning

### DIFF
--- a/ehr/resources/data/species.tsv
+++ b/ehr/resources/data/species.tsv
@@ -1,10 +1,10 @@
-common	scientific_name	id_prefix	mhc_prefix	blood_per_kg	max_draw_pct	blood_draw_interval
-Baboon				60.0	0.2	30.0
-Cotton-top Tamarin	Saguinus oedipus	so	Saoe	60.0	0.2	30.0
-Cynomolgus	Macaca fascicularis	cy	Mafa	60.0	0.2	30.0
-Pigtail	Macaca Nemestrina		Mane	60.0	0.2	30.0
-Rhesus	Macaca mulatta	r|rh	Mamu	60.0	0.2	30.0
-Sooty Mangabey	Cercocebus atys		Ceat	60.0	0.2	30.0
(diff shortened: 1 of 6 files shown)


bbimber:
My top-level comment is that I worry this is ballooning into something that should be most appropriately handled in the WNPRC module. I cant argue that any one of these changes by itself is unreasonable, nor is this feature inherently unreasonable, but it is starting to add more and more to a process that is already not exactly simple. 

Maybe a better idea would be to expose a mechanism to allow modules to register a BloodDrawCalculatorFactory class or something to that effect? We could have a default implementation (the existing code), but a module could register something to override this? That interface would probably be pretty simple, needing:
```
interface BloodDrawCalculatorFactory {
  boolean isAvailable(Container c, User u);

  BloodDrawCalculator create(Container c, User u, ....whatever other state variables we need...);
}

interface BloodDrawCalculator {
  void doCalculation(.....whatever set of variables are required to separate this....it can also add errors to the error array directly...);
}
```
I am suggesting separating the factory and BloodDrawCalculator because this would let the code most cleanly create it once per transaction.

csebranek:
@bbimber I can see your point about refactoring some of this to make it easier for other modules to customize it, I am thinking a separate PR might be good for the bulk of that work.

labkey-martyp (approved):
@csebranek I went over this with Daniel this week. The column itself is general enough to add to the table. Then in a follow up PR do something like what Ben is describing to allow customization. Happy to discuss approaches there. And if you can show this in the next EHR collaboration in case there are other groups who want something similar. 

I tested the upgrade scripts. This looks good.

PR #821, opened 2024-09-03
Augment kinship calculation to check modification date and only run as needed

bbimber:
The EHR has a longstanding pipeline job that calculates and stores kinship and inbreeding. For a large colony, the compute time for this R job is pretty big, and it requires a long time to import the results. This job historically runs once/day as a scheduled pipeline job. 

The underlying data does not necessarily change every day, and therefore it's not actually necessarily to run it daily. This PR as opt-in behavior such that:

1) If the study.pedigree query (which is typically defined in the center-specific module) contains as column named modified, the new behavior starts. If the source query does not have modified, nothing changes.

2) If study.pedigree contains modified, then the scheduler will first query the max(modified). It compares this against the last run time of the job. If this value is less than the job's last run time, the scheduler aborts and doesnt submit that unnecessary job. 

Again, unless the center chooses to add 'modified' to their study.pedigree query, nothing changes.

### DIFF
--- a/ehr/resources/queries/study/demographicsParents.sql
+++ b/ehr/resources/queries/study/demographicsParents.sql
@@ -23,7 +23,8 @@ SELECT
         WHEN d.dam IS NOT NULL AND d.sire IS NOT NULL THEN 2
         WHEN d.dam IS NOT NULL OR d.sire IS NOT NULL THEN 1
         ELSE 0
-        END as numParents
+        END as numParents,
+    d.modified
 FROM study.demographics d
 
 UNION ALL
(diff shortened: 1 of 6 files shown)


bbimber:
I believe that NIRC data import test failure is unrelated to this. that test has been failing for a while: https://teamcity.labkey.org/test/-6635263291310752578?currentProjectId=LabKey_247Release_External_Ehr&expandTestHistoryChartSection=true

labkey-martyp (approved):
Thanks @bbimber, this looks like a good feature. I'll bring this up in the next EHR collab so folks know about it.

bbimber:
thanks @labkey-martyp

PR #833, opened 2024-09-26
Adjustments to blood summary plot

csebranek:
#### Rationale
Adding a legend for readability, and starting the y-axis at zero to more accurately show if an animal is almost out of blood.

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/ehr/resources/web/ehr/panel/BloodSummaryPanel.js
+++ b/ehr/resources/web/ehr/panel/BloodSummaryPanel.js
@@ -309,6 +309,15 @@ Ext4.define('EHR.panel.BloodSummaryPanel', {
             return text;
         });
     },
+    getMaxBloodAvailValue: function(rows){
+
+        var allowableBloodVals = [];
+        for (var i = 0; i < rows.length; i++){
+            allowableBloodVals.push(rows[i].allowableDisplay.value);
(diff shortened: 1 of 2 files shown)


labkey-martyp (commented):
This looks great. As discussed, a couple changes for the y axis,
1. Handle the case where available blood goes below zero. Should just need to adjust the range and the plot will automatically shade that in red. Could consider extending y axis below zero if getting close to zero. Like subtracting 10% of max allowable from bottom is negative, then that is the y axis lower range, otherwise it's zero.
2. Maybe just adding 10% instead of 10mL to the top range of the y axis.

labkey-martyp (approved):
@csebranek This looks good. Please let me know when you're ready to merge this in. I have another PR that will need to be merged in as well.

bbimber:
@csebranek and @labkey-martyp: out of curiosity, why does EHR need a separate server-side action to get the blood data? Why doesnt select rows with very minor client-side reformatting accomplish the same thing?

labkey-martyp:
@bbimber Yeah I don't see anything that couldn't be done with select rows. I don't remember the reason we gave it it's own action. @csebranek something we can consider if there are further modifications here.

PR #861, opened 2024-11-15
Fix incorrect R syntax

bbimber:
This includes two minor improvements:

1) I noticed the R script in the EHR (which is supposed to document what dependencies are needed), doesnt have valid R code. This is never directly executed, but it might as well be correct.

2) This adds decimal formatting to the output of the kinship job. Not exactly mission-critical, but a minor improvement that makes it a little easier to see progress.

### DIFF
--- a/ehr/resources/pipeline/install.r
+++ b/ehr/resources/pipeline/install.r
@@ -5,4 +5,4 @@
 ##
 
 #installs required packages for populateKinship.r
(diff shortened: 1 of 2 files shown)


bbimber:
I am assuming those test failures are not related to this...

labkey-martyp (approved):
Thanks for fixing/improving those.

labkey-martyp:
> I am assuming those test failures are not related to this...

Yes those test failures are unrelated. This can be merged in. Thanks for checking.

PR #871, opened 2024-12-08
Set caseid on all sections in surgery form

labkey-martyp:
#### Rationale
If a section in the surgery form contains a caseid column, set that column value when selecting/creating a case. Assumes forms have one case per id. Can override caseUpdateStores if a different set of stores to be updated.

#### Related Pull Requests
* TBD

#### Changes
* Loop through stores and find records for the animal in the case, updating caseid on the records if the column exists.

### DIFF
--- a/ehr/resources/web/ehr/window/OpenSurgeryCasesWindow.js
+++ b/ehr/resources/web/ehr/window/OpenSurgeryCasesWindow.js
@@ -268,6 +268,16 @@ Ext4.define('EHR.window.OpenSurgeryCasesWindow', {
                                     console.log('updating procedure with caseid');
                                     rec.set('caseid', row.objectid);
                                 }, this);
+
+                                this.caseUpdateStores.forEach(function(store){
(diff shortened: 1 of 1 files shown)


labkey-bpatel (approved):

ehr/resources/web/ehr/window/OpenSurgeryCasesWindow.js
```diff
@@ -336,7 +356,8 @@ EHR.DataEntryUtils.registerDataEntryFormButton('OPENSURGERYCASES', {
         }
 
         Ext4.create('EHR.window.OpenSurgeryCasesWindow', {
-            sourceStore: clientStore
+            sourceStore: clientStore,
+            caseUpdateStores: panel.storeCollection.clientStores.items,
```
bbimber, line 360: This is picky, but wouldnt it be a little more normal to make a pointer to the StoreCollection, rather than 'items', which is more like an internal field of that class? Either accomplishes your goal, obviously.

PR #899, opened 2025-03-02
Formulary rounding fixes

labkey-martyp:
#### Rationale
A few fixes for formulary rounding

#### Changes
* Fix conditional when amount rounding should be done
* Update rounding to nearest function to handle floating point precision issues.
* Allow greater precision in rounding field

### DIFF
--- a/ehr/resources/web/ehr/utils.js
+++ b/ehr/resources/web/ehr/utils.js
@@ -497,19 +497,9 @@ EHR.Utils = new function(){
                 return val;
             }
 
-            var remainder = val % round;
-            var remainderPct = remainder / round;
-            var base = Math.floor(val / round);
-            if (remainder === 0){
-                return val;
-            }
-            //note: JS seems to handle division poorly in situations like 4 / 0.1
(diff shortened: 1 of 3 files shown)


ehr/resources/web/ehr/utils.js
```diff
@@ -497,19 +497,9 @@ EHR.Utils = new function(){
...
-            //note: JS seems to handle division poorly in situations like 4 / 0.1
-            else if (remainderPct < 0.5 || remainderPct > 0.9999){
-                return (base * round);
-            }
-            else {
-                return (base * round) + round;
-            }
+            let rounded = Math.round(val / round) * round;
+            let precision = (round.toString().split('.')[1] || "").length; // Get decimal places in `round`
+            return rounded.toFixed(precision); // toFixed handles floating point precision issues
```
labkey-bpatel, line 502: nice, good to know about toFixed().

labkey-bpatel (approved):

ehr/resources/web/ehr/utils.js
labkey-martyp: floating point is the worst

PR #915, opened 2025-03-19
Misc code cleanup; add vet to users and groups in tests

labkey-jeckels:
#### Rationale
Add a new vet user and group as part of automated test setup. Clear many IntelliJ warnings.

#### Changes
* Clean up generics
* Misc other code cleanup
* Add vet user and group
* Include both site and project groups when populating the `ehr_lookups.veterinarian` table

### DIFF
--- a/ehr/src/org/labkey/ehr/query/EHRLookupsUserSchema.java
+++ b/ehr/src/org/labkey/ehr/query/EHRLookupsUserSchema.java
@@ -36,9 +36,7 @@ import org.labkey.api.ldk.table.ContainerScopedTable;
 import org.labkey.api.query.FieldKey;
 import org.labkey.api.query.FilteredTable;
 import org.labkey.api.query.QueryForeignKey;
-import org.labkey.api.query.QueryService;
 import org.labkey.api.query.SimpleUserSchema;
-import org.labkey.api.query.UserSchema;
 import org.labkey.api.security.Group;
 import org.labkey.api.security.MemberType;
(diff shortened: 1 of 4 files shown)


ehr/src/org/labkey/ehr/query/EHRLookupsUserSchema.java
```diff
@@ -271,9 +266,9 @@ private TableInfo createVeterinariansTable(String name, ContainerFilter cf)
                 else
                 {
                     Group assignedGroup = SecurityManager.getGroup(r.getUserId());
-                    if (assignedGroup != null && !assignedGroup.isProjectGroup())
+                    if (assignedGroup != null)
```
labkey-martyp, line 269: Not really sure what the original use case was here. I don't see anyone using vets that wouldn't be listed in EHR or having different permissions at project level.

labkey-martyp (approved):

ehr/src/org/labkey/ehr/query/EHRLookupsUserSchema.java
bbimber: I assume I wrote that, but it's been too long to recall. I cannot think of any reason to differentiate between site groups and project-level groups. If no center is actually using a project-level group in this table then it's a moot point and probably doesnt need to exist.

Given how long ago this was, it's equally likely that some long-ago quirk the in SecurityManager API required this kind of check for some reason.

PR #918, opened 2025-03-22
codeowners fix

labkey-martyp:
#### Rationale
CODEOWNERS not working, add wildcard.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-@labkey-martyp
+*	@labkey-martyp


.github/CODEOWNERS
```diff
@@ -1 +1 @@
-@labkey-martyp
\ No newline at end of file
+*	@labkey-martyp
```
labkey-bpatel, line 1: Is that a tab space in between?

labkey-bpatel (approved):

.github/CODEOWNERS
labkey-martyp: yes

PR #937, opened 2025-04-10
DatabaseIdentifier

labkey-matthewb:
#### Rationale
Generating correct sql is very dependent on everyone writing identifier consistently.  This is often handled by using append(SqlDialect.makeLegalIdentifier(col.getAlias())) or luck append(col.getName()) or append(col.getAlias()).

The idea is to pass around database identifiers such as table names, column names, both generated and from db metadata, using a class DatabaseIdentifier.

#### Related Pull Requests
- 

#### Changes
-


labkey-martyp (approved):
I did your refactor in both places

labkey-adam (approved):

PR #949, opened 2025-05-01
User dropdowns should only show active users

labkey-martyp:
#### Rationale
All assignment dropdowns are showing inactive users along with active users. I'm not aware of any cases that a review, task, request, etc would be assigned to an inactive user.

This does not affect historical data.

#### Changes
* Add user active status to PrincipalsWithoutAdmin
* Update UsersAndGroupsCombo filter

### DIFF
--- a/ehr/resources/queries/core/PrincipalsWithoutAdmin.sql
+++ b/ehr/resources/queries/core/PrincipalsWithoutAdmin.sql
@@ -19,7 +19,8 @@ SELECT
   u.DisplayName,
   'u' as Type,
   u.FirstName,
-  u.LastName
+  u.LastName,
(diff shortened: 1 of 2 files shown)


labkey-jeckels:
@labkey-martyp will this work in practice, given your question about the permissions needed to see/use the `Active field?

labkey-martyp:
This does not work as-is. Working on a refactor.

labkey-martyp:
This PR now works

labkey-bpatel (approved):
Verified that the 'PrincipalsWithoutAdmin' query is accessible for the non-admins, which ensures that the Active users filter works as intended during data entry.

PR #974, opened 2025-06-24
Issue 52277: SNDTest.testSuperPackageApis fails with DB deadlock

labkey-jeckels:
#### Rationale
This test is failing intermittently but regularly. It's a SQLServer deadlock between loading the schema browser and doing the import. We can sidestep it by waiting for the query dependency analysis to complete before hitting the APIs.

#### Changes
* Wait for the Dependency Report
* Just reload the page instead of going back through the schema browser after each change
* Minor cleanup


snd/test/src/org/labkey/test/tests/snd/SNDTest.java
```diff
@@ -1167,8 +1168,8 @@ public void testSNDImport()
         _fileBrowserHelper.checkFileBrowserFileCheckbox(REMOVE_ALL_ATTRIBUTES_FILE);
         _fileBrowserHelper.selectImportDataAction("SND document import");
         waitForPipelineJobsToComplete(++EXPECTED_IMPORT_JOBS, "SND Import ("+REMOVE_ALL_ATTRIBUTES_FILE+")", true, IMPORT_WAIT_TIME);
-
-        checkExpectedErrors(1);
+            resetErrors();
+//        checkExpectedErrors(1);
```
labkey-danield: Maybe remove this commented line? Or add a comment as to why it is commented out.

labkey-danield (approved):

snd/test/src/org/labkey/test/tests/snd/SNDTest.java
labkey-jeckels: Thanks for the reminder to revert that edit. I had to do that hack to run the test locally but didn't intend to push it.

PR #975, opened 2025-06-25
Better handling of GeneticCalculationsImportTask cancels

bbimber:
If the user hits cancel on a pipeline job, they should reasonably expect the job to cancel. While GeneticCalculationsImportTask already has a check for job.isCancelled(), I dont think isCancelled() actually returns the correct value unless job.updateStatusForTask() is called. This PR adds this, which makes the job check the current status of the job every time it completes a batch of imports, and will therefore cancel the job in a more timely manner. 

While I was editing this, I also made an option to specify dayOfWeek, in addition to hourOfDay for the cron schedule. If blank the job will run nightly, as-is. The would allow admins to pick a day of the week to convert this into a weekly job.


labkey-martyp (approved):
Looks good.

PR #997, opened 2025-07-28
EHR Link Crawler Test

labkey-martyp:
#### Rationale
Core EHR test that crawls links in the EHR folder or subfolders to check for broken links. Help find CSP violations and provide a new test to ensure links expected to work do not have regressions.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/2585
* https://github.com/LabKey/johnsHopkinsEHRModules/pull/494
* https://github.com/LabKey/nircEHRModules/pull/559
* https://github.com/LabKey/snprcEHRModules/pull/835
* https://github.com/LabKey/tnprcEHRModules/pull/131
* https://github.com/LabKey/wnprc-modules/pull/855
* https://github.com/LabKey/onprcEHRModules/pull/1402

#### Changes
* Add testCrawlEhrLinks
* Add skip links known to cause trouble in test setups
* Update EHRApp test skip links

### DIFF
--- a/ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
+++ b/ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
@@ -391,6 +391,7 @@ public abstract class AbstractGenericEHRTest extends AbstractEHRTest
 
             if (clickable)
             {
+                waitFor(() -> (getDriver().getCurrentUrl() != null && !getDriver().getCurrentUrl().equalsIgnoreCase("about:blank")), WAIT_FOR_JAVASCRIPT);
                 URL url = getURL();
                 assertFalse("URL " + url + " is empty.", isPageEmpty());
                 assertNoLabKeyErrors();
--- a/ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
+++ b/ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
@@ -346,7 +346,8 @@ public abstract class AbstractGenericEHRTest extends AbstractEHRTest
         String href = anchor.getDomAttribute("href");
         if (href != null && !href.startsWith("#"))
         {
-            if (skipLinksForValidation().stream().anyMatch(s -> href.toLowerCase().contains(s.toLowerCase())))
+            String decodedHref = EscapeUtil.decodeUriPath(href);
+            if (skipLinksForValidation().stream().anyMatch(s -> decodedHref.toLowerCase().contains(s.toLowerCase())))
             {
                 log(href + " is specified as an exception to link validation. Skipping validation.");
                 return validUrl;
@@ -367,8 +368,8 @@ public abstract class AbstractGenericEHRTest extends AbstractEHRTest
                 // not a full URL so not external. Carry on.
             }
 
-            // scope this to ehr folder and subfolders
-            if (!EscapeUtil.decodeUriPath(href).contains(getContainerPath()))
+            // scope this to admin, ehr folder and subfolders
+            if (!decodedHref.contains(getContainerPath()) && !decodedHref.startsWith("/admin"))
             {
                 log(href + " is in a different folder than the EHR folder, " + getContainerPath() + ". Skipping validation.");
                 return validUrl;
--- a/ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
+++ b/ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
@@ -30,7 +30,9 @@ import org.labkey.test.util.LoggedParam;
 import org.labkey.test.util.PortalHelper;
 import org.labkey.test.util.ext4cmp.Ext4ComboRef;
 import org.labkey.test.util.external.labModules.LabModuleHelper;
+import org.labkey.test.util.selenium.WebDriverUtils;
 import org.openqa.selenium.By;
+import org.openqa.selenium.Keys;
 import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.WebElement;
 
@@ -307,6 +309,21 @@ public abstract class AbstractGenericEHRTest extends AbstractEHRTest
         resetErrors();  //note: inserting records without permission will log errors by design.  the UI should prevent this from happening, so we want to be aware if it does occur
     }
 
+    // Clicks the link and switches to the window if it's a viable link. Otherwise throws an exception.
+    private void verifyLInk(WebElement link)
+    {
+        int winCount = getDriver().getWindowHandles().size();
+        link.sendKeys(Keys.chord(WebDriverUtils.MODIFIER_KEY, Keys.ENTER));
+
+        // Short wait for a new window to open. If not then throw exception
+        boolean winOpen = waitFor(() -> getDriver().getWindowHandles().size() > winCount, 1000);
+        if (!winOpen)
+            throw new IllegalStateException("Link did not open new window in tab.");
+
+        List<String> windows = new ArrayList<>(getDriver().getWindowHandles());
+        getDriver().switchTo().window(windows.get(1));
+    }
+
     protected List<String> skipLinksForValidation()
     {
         return List.of(
@@ -385,7 +402,7 @@ public abstract class AbstractGenericEHRTest extends AbstractEHRTest
 
         try
         {
-            openLinkInNewWindowOrThrow(anchor);
+            verifyLInk(anchor);
         }
         catch (WebDriverException | IllegalStateException e)
         {
--- a/ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
+++ b/ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
@@ -382,7 +382,7 @@ public abstract class AbstractGenericEHRTest extends AbstractEHRTest
 
             try
             {
-                openLinkInNewWindow(anchor);
+                openLinkInNewWindowOrThrow(anchor);
             }
             catch (WebDriverException | IllegalStateException e)
             {
--- a/ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
+++ b/ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
@@ -391,14 +391,20 @@ public abstract class AbstractGenericEHRTest extends AbstractEHRTest
 
             if (clickable)
             {
-                waitFor(() -> (getDriver().getCurrentUrl() != null && !getDriver().getCurrentUrl().equalsIgnoreCase("about:blank")), WAIT_FOR_JAVASCRIPT);
-                URL url = getURL();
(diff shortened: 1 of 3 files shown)


ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
```diff
@@ -300,6 +306,160 @@ private void testUserAgainstAllStates(@LoggedParam EHRUser user)
...
+        String validUrl = null;
+
+        String href = anchor.getDomAttribute("href");
+        if (href != null && !href.startsWith("#"))
+        {
+            String decodedHref = EscapeUtil.decodeUriPath(href);
+            if (skipLinksForValidation().stream().anyMatch(s -> decodedHref.toLowerCase().contains(s.toLowerCase())))
+            {
+                log(href + " is specified as an exception to link validation. Skipping validation.");
+                return validUrl;
```
labkey-tchad: `validUrl` will always be `null` here. Returning `validUrl` adds unnecessary cognitive load to anybody reading this method.
```suggestion
                return null;
```

ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
```diff
@@ -300,6 +306,160 @@ private void testUserAgainstAllStates(@LoggedParam EHRUser user)
...
+            }
+
+            // Ensure link is not external
+            try
+            {
+                URL url = new URL(href);
+                if (!url.getHost().equalsIgnoreCase(getURL().getHost()))
+                {
+                    log(href + " is an external link. Skipping validation.");
+                    return validUrl;
```
labkey-tchad:
```suggestion
                    log(href + " is an external link. Skipping validation.");
                    return null;
```

ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
```diff
@@ -300,6 +306,160 @@ private void testUserAgainstAllStates(@LoggedParam EHRUser user)
...
+            catch (MalformedURLException e)
+            {
+                // not a full URL so not external. Carry on.
+            }
+
+            // scope this to admin, ehr folder and subfolders
+            if (!decodedHref.contains(getContainerPath()) && !decodedHref.startsWith("/admin"))
+            {
+                log(href + " is in a different folder than the EHR folder, " + getContainerPath() + ". Skipping validation.");
+                return validUrl;
```
labkey-tchad:
```suggestion
                log(href + " is in a different folder than the EHR folder, " + getContainerPath() + ". Skipping validation.");
                return null;
```

ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
```diff
@@ -300,6 +306,160 @@ private void testUserAgainstAllStates(@LoggedParam EHRUser user)
...
+                    return validUrl;
+                }
+            }
+            catch (MalformedURLException e)
+            {
+                // not a full URL so not external. Carry on.
+            }
+
+            // scope this to admin, ehr folder and subfolders
+            if (!decodedHref.contains(getContainerPath()) && !decodedHref.startsWith("/admin"))
```
labkey-tchad: `Crawler.ControllerActionId` will parse out the controller, action, and containerPath of a URL (`new Crawler.ControllerActionId(href)`). Note that it does not include a leading slash like `getContainerPath()` does.

labkey-tchad: Why does this crawl admin actions? Our regular crawler should definitely hit those.

ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
```diff
@@ -300,6 +306,160 @@ private void testUserAgainstAllStates(@LoggedParam EHRUser user)
...
+            {
+                log(href + " is in a different folder than the EHR folder, " + getContainerPath() + ". Skipping validation.");
+                return validUrl;
+            }
+        }
+
+        if (anchor.isDisplayed() && anchor.isEnabled())
```
labkey-tchad: Links can't be disabled ([it isn't a valid attribute for anchor tags](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/disabled)).
`anchor.isEnabled()` will always return true.
```suggestion
        if (!anchor.isDisplayed())
            return null;
```

ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
```diff
@@ -300,6 +306,160 @@ private void testUserAgainstAllStates(@LoggedParam EHRUser user)
...
+                assertFalse("URL " + url + " is empty.", isPageEmpty());
+                assertNoLabKeyErrors();
+
+                // assertNoLabKeyErrors does not catch all types of errors
+                assertElementNotPresent("LabKey error found for URL " + url, Locators.labkeyErrorHeading);
+
+                // record link as valid and cleanup
+                validUrl = url.toString();
+                switchToWindow(0);
+                quietlyCloseExtraWindows();
```
labkey-tchad: No need to use a special method to close all of the windows. There's only one to close.
```suggestion
                // record link as valid and cleanup
                getDriver().close();
                switchToWindow(0);
                return url.toString();
```

ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
```diff
@@ -300,6 +306,160 @@ private void testUserAgainstAllStates(@LoggedParam EHRUser user)
...
+
+        log(anchors.size() + " possible links found.");
+        int validatedCount = 0;
+        Set<String> validLinksOnPage = new HashSet<>();
+        for (WebElement anchor : anchors)
+        {
+            // Only validate links once
+            String href = anchor.getDomAttribute("href");
+            if (href != null && validLinksOnPage.contains(href))
+                continue;
```
labkey-tchad: Skip if link has already been crawled?
```suggestion
            if (href != null && validLinksOnPage.contains(href) || crawledLinks.contains(href))
                continue;
```

ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
```diff
@@ -300,6 +306,160 @@ private void testUserAgainstAllStates(@LoggedParam EHRUser user)
...
+            }
+        }
+
+        if (anchor.isDisplayed() && anchor.isEnabled())
+        {
+            clickable = true;
+
+            try
+            {
+                openLinkInNewWindowOrThrow(anchor);
```
labkey-tchad: What sort of links don't open successfully? This seems like an error state.
If there are links that we expect to not open, we should filter them out earlier.
Waiting for a thing that won't happen can waste a lot of time (`switchToWindow` waits ten seconds for the new tab to appear)

Opening these links at all in these phase seems inefficient. `validatePageLinks` also navigates to the links, so we're hitting each link twice.

ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
labkey-martyp: Nice I can get rid of quietlyCloseExtraWindows

ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
labkey-martyp: Got it

ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
labkey-martyp: I meant to just validate that links in the EHR pointing to admin are valid URLs. I just added admin to the default urls to skip crawling, but it will continue to validate the links to there from the EHR.

labkey-martyp: Updated

ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
labkey-martyp: good catch thanks

ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
labkey-martyp: So our pages have lots of anchor tags on them that aren't links, and the ones that are links can have a href, onclick handler or click listener added a variety of ways in JS. Instead of trying to pre-determine which anchor tags are actual links, this clicks on the anchor tags and sees if a new window opens. That's why I added the throw immediately in openLinkInNewWindow, to not wait ten seconds for a new window.

We are hitting each link twice. I'm trying to do somewhat of a breadth first search validating the links on one page at a time (easier logging and troubleshooting) and iterating through the anchor tags to see which are actual links. Once determined if its an actual link then the URL gets queued up for crawling. I don't think it will work to queue up a bunch of anchor tag webelements then navigate away. It needs to determine what anchor tags are actual links for crawling. I'm seeing this test taking 7-12 min. A little longer but it's pretty comprehensive.

labkey-martyp: Skipping the dataset view is making these run a couple minutes faster.

ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
```diff
@@ -300,6 +307,162 @@ private void testUserAgainstAllStates(@LoggedParam EHRUser user)
...
+                "junit-begin.view",
+                "admin-",
+                "ehr-datasets.view"
+        );
+    }
+
+    private String validLink(WebElement anchor)
+    {
+        String href = anchor.getDomAttribute("href");
+        if (href != null && !href.startsWith("#") && !href.equalsIgnoreCase("undefined"))
```
labkey-tchad, line 364: `<a href="undefined">` definitely seems like an error case. Maybe the `NightlyTestServer` module property needs to be set or [ehrBegin.html](https://github.com/LabKey/wnprc-modules/blob/eefd9f45105ec1f97aca5332b86be06854770c60/WNPRC_EHR/resources/views/ehrBegin.html#L127) should handle it being unset?

labkey-tchad (approved):

ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
labkey-martyp: I did update the WNPRC test to set NightlyTestServer, unfortunately it's not the only place where this happens. The issue here is Crawler.ControllerActionId doesn't handle this and throws a low level exception (NPE or something can't remember). This will still throw an error just doesn't try to parse apart the href.

PR #1003, opened 2025-08-05
Housing overlaps form changes

kollil:
Modifying the housing overlaps selection form in ehr.
**Form url:**
https://prime.ohsu.edu/ehr/ONPRC/EHR/housingOverlaps.view?groupById=1

Functionality changes:
- Add Area field with multi select to pick rooms
- Add Building field with multi select to pick rooms
- Update to handle multi-room selection


labkey-martyp:
@kollil Hi Lakshmi, can you please try this out and see if it works for your usage. I did add a buildings field as well to make it more universal for the EHR, but we could hide that with a url parameter if necessary. Thanks.

labkey-martyp (approved):

PR #1023, opened 2025-08-28
Use "Id" for insert rows key instead of "participantid" (which is a column alias)

cnathe:
#### Rationale
The related PR updated which column alias mapping is generated during the insert rows call made for the GeneticCalculationsImportTask dataset data. This PR updates the data row keys to use the column name instead of the alias.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6939/files#diff-2a61025ca8763e384dc685f377df1b80ae666e6d01f5182f2b10abfc7c894accL252

#### Changes
- import dataset row data using "Id" as the subject key

### DIFF
--- a/ehr/src/org/labkey/ehr/pipeline/GeneticCalculationsImportTask.java
+++ b/ehr/src/org/labkey/ehr/pipeline/GeneticCalculationsImportTask.java
@@ -690,7 +690,6 @@ public class GeneticCalculationsImportTask extends PipelineJob.Task<GeneticCalcu
                 }
 
                 row.put("Id", subjectId);
-                //row.put("participantid", subjectId);
                 row.put("date", date);
                 row.put("coefficient", Double.parseDouble(fields[1]));
 
--- a/ehr/src/org/labkey/ehr/pipeline/GeneticCalculationsImportTask.java
+++ b/ehr/src/org/labkey/ehr/pipeline/GeneticCalculationsImportTask.java
@@ -689,8 +689,8 @@ public class GeneticCalculationsImportTask extends PipelineJob.Task<GeneticCalcu
                     continue;
                 }
 
-                //row.put("Id", subjectId);
-                row.put("participantid", subjectId);
+                row.put("Id", subjectId);
+                //row.put("participantid", subjectId);
                 row.put("date", date);
                 row.put("coefficient", Double.parseDouble(fields[1]));
 


ehr/src/org/labkey/ehr/pipeline/GeneticCalculationsImportTask.java
```diff
@@ -689,8 +689,8 @@ private static void processInbreeding(Container c, User u, File pipelineDir, Log
                     continue;
                 }
 
-                //row.put("Id", subjectId);
-                row.put("participantid", subjectId);
+                row.put("Id", subjectId);
```
bbimber, line 692: @cnathe, this might be a very old LK bug that has since been fixed, but this type of thing exists b/c labkey used to require "participantid".  If that's no longer true, great. 

However, is there any reason to keep the commented out line?

bbimber (changes requested):

ehr/src/org/labkey/ehr/pipeline/GeneticCalculationsImportTask.java
labkey-jeckels: There are many other places in the EHR that assume "Id" as the column name, but we could easily use the configured name here:

StudyService.get().getSubjectColumnName(c)

labkey-jeckels: > @cnathe, this might be a very old LK bug that has since been fixed, but this type of thing exists b/c labkey used to require "participantid". If that's no longer true, great.
> 
> However, is there any reason to keep the commented out line?

Maybe I'm confused about what you mean, but there's no commented out line after this change. Did you want to retain it?

labkey-jeckels (approved):

ehr/src/org/labkey/ehr/pipeline/GeneticCalculationsImportTask.java
```diff
@@ -689,8 +689,8 @@ private static void processInbreeding(Container c, User u, File pipelineDir, Log
...
+                //row.put("participantid", subjectId);
```
bbimber: @labkey-jeckels: isnt this commented out?

labkey-jeckels: Ha, yes. Sorry. Your comment pushed it below the visible section of the page when I was looking at the diff.

ehr/src/org/labkey/ehr/pipeline/GeneticCalculationsImportTask.java
bbimber: the view github creates here is confusing. see line 693 for the commented one.

I do not have strong views on hard-coding names. EHR does hard-code "Id" in a lot of places such that it's not especially practical to eliminate. Even if we replace the java code, there's a lot of JS/HTML/SQL. "participantid" is not used in EHR, but you will find several places in java insert code that it is (or at least was) used, because LK used to require it. I dont recall why, but it was a weird LK quirk. That was the reason I posted about this.

ehr/src/org/labkey/ehr/pipeline/GeneticCalculationsImportTask.java
cnathe: I've removed the commented line re: particpiantid

bbimber (approved):

PR #1018, opened 2025-08-21
Issue 53225: Unable to delete row from ehr_lookups.procedures

labkey-bpatel:
#### Rationale
Issue [53225](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53225): Unable to delete row from ehr_lookups.procedures

The failure occured in LookupValidationHelper.verifyNotUsed() for EHR clients that did not have study.encounters table in their dataset.

#### Related Pull Requests
* https://github.com/LabKey/LabDevKitModules/pull/259
* https://github.com/LabKey/nircEHRModules/pull/579
* https://github.com/LabKey/johnsHopkinsEHRModules/pull/509

#### Changes
* Update procedures.js trigger to use registerHandlerForQuery(), enabling it to be unregistered for EHR clients without a study.encounters table.
* Update trigger scripts using LookupValidationHelper.verifyNotUsed() to add the script name to error messages (making it easier to trace the source since trigger failures often return unhelpful messages).

PR #1029, opened 2025-09-01
Use EHR Study Container Entity Id instead of Study Entity Id to generate cache key

labkey-bpatel:
#### Rationale
Generating a cache key with the Study Entity Id requires the Study to be loaded. This becomes an issue when populating lookups - ex. for ehr_lookups.procedures, we want to allow procedures trigger to be registered and unregistered (as done in the related PR), which hits the code path addressed in this PR. In tests, lookups are populated before the Study is created, so the Study Entity Id isn't always available at that point - this causes test failures for the clients that populate ehr_lookups.procedures.

#### Related Pull Requests
* https://github.com/LabKey/ehrModules/pull/1018

#### Changes
* Use EHR Study Container Entity Id instead of Study Entity Id to generate cache key

### DIFF
--- a/ehr/src/org/labkey/ehr/security/EHRSecurityManager.java
+++ b/ehr/src/org/labkey/ehr/security/EHRSecurityManager.java
@@ -224,29 +224,25 @@ public class EHRSecurityManager
         if (targetContainer == null)
             return Collections.emptyMap();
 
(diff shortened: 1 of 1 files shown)


bbimber (approved):
i cant think of any reason this would be a problem, but I didnt do any testing.

labkey-martyp (approved):

PR #1045, opened 2025-09-22
Don't crawl any study links

labkey-martyp:
#### Rationale
No reason to crawl study links for EHR tests. Add those to skip list.

#### Changes
* Add study to skip list


labkey-chrisj (approved):
Makes sense to me

PR #1052, opened 2025-09-30
Reference security-category table columns by header text

labkey-chrisj:
#### Rationale
This change attempts to address intermittent test failures in [SNDTest.categoryPermissionsUI](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_EhrSqlserver/3678823?buildTab=tests&name=org.labkey.test.tests.snd.SNDTest.categoryPermissionsUI&expandedTest=build%3A%28id%3A3678823%29%2Cid%3A2000000596).

The theory of this change is that the order of the columns does not appear to be fixed, but the test references (and asserts) based on hard-coded column indexes.  This adds a method in the test to get the column index based on the text of the column header.

### DIFF
--- a/snd/test/src/org/labkey/test/tests/snd/SNDTest.java
+++ b/snd/test/src/org/labkey/test/tests/snd/SNDTest.java
@@ -2125,7 +2125,13 @@ public class SNDTest extends BaseWebDriverTest implements SqlserverOnlyTest
     private int getPermissionTableColumnIndex(String column)
     {
         var headersLoc = Locator.id("category-security").descendant(Locator.tagWithClass("th", "group-hdr"));
-        return getTexts(headersLoc.findElements(getDriver())).indexOf(column) +1;
+        var columnHeaders =  getTexts(headersLoc.findElements(getDriver()));
+        checker().withScreenshot("column_not_found").fatal()
+                .wrapAssertion(()-> Assertions.assertThat(columnHeaders)
+                .as(String.format("Column %s not found", column))
+                .contains(column));
+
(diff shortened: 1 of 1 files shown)


labkey-martyp (approved):
The change looks good to me. I kicked off a few more runs on TC to test flakiness.

labkey-chrisj:
Here's a link to my run in this branch: https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_EhrSqlserver?branch=category_security_table_columns

labkey-martyp:
I also ran just the SND suite a few more times and they all passed. https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_LabkeyPremiumTrunk_GitModules_SndSqlserver?branch=category_security_table_columns&buildTypeTab=overview

snd/test/src/org/labkey/test/tests/snd/SNDTest.java
```diff
@@ -2121,6 +2122,12 @@ private void runTestsInAPIFrameWork()
         assertTextPresent("Complete","Failed tests: 0");
     }
 
+    private int getPermissionTableColumnIndex(String column)
+    {
+        var headersLoc = Locator.id("category-security").descendant(Locator.tagWithClass("th", "group-hdr"));
+        return getTexts(headersLoc.findElements(getDriver())).indexOf(column) +1;
```
labkey-tchad: Consider throwing an exception if the column doesn't exist

labkey-tchad (approved):
Thank you for doing this

PR #1070, opened 2025-11-25
EHR study participant override

labkey-martyp:
#### Rationale
There are still a couple places in EHR, such as site search results, that show the base study participant view. This PR overrides that view to use EHR participant view. Goes to base EHR participantView but this action should be overridden to custom participantViews as necessary.

#### Changes
* participant HTML override with redirect

### DIFF
--- a/ehr/resources/views/participant.html
+++ b/ehr/resources/views/participant.html
@@ -1,0 +1,6 @@
+
+<script nonce="<%=scriptNonce%>">
+    const id = LABKEY.ActionURL.getParameter("participantId");
+    const url = LABKEY.ActionURL.buildURL("ehr", "participantView", null, {participantId: id})
+    window.location.href = url;
+</script>


bbimber:
@labkey-martyp: I'm not sure of the background, but is there any reason not to do this as a java action in EhrController? That would be a cleaner redirect without needing to load anything on the client, just to immediately redirect and load something else. 

If there is a need for center-specific modules to be able to override it, could that java code check whether any active module has a participant.html file and defer to that if present?

labkey-martyp:
@bbimber Unfortunately, the study module is looking specifically for a custom module html view to override the study's participant view. An action in EHR controller will not override the base study participant view. There's a ticket open and some discussion to allow an action override in study configuration of the participant view. This is a little clunky but gets the basic functionality for a fairly rare use case.

bbimber:
OK

labkey-bpatel (approved):

PR #1073, opened 2025-12-12
Fix for testCrawlEhrLinks() failure

labkey-bpatel:
#### Rationale
testCrawlEhrLinks() failure
https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_EhrPostgres/3766798

org.openqa.selenium.StaleElementReferenceException
...
 at app//org.labkey.test.tests.ehr.AbstractGenericEHRTest.validatePageLinks(AbstractGenericEHRTest.java:448)
 at app//org.labkey.test.tests.ehr.AbstractGenericEHRTest.testCrawlEhrLinks(AbstractGenericEHRTest.java:480)

The links are probably still getting loaded when the page is rendered and updating the DOM causing StaleElementReferenceException.

#### Related Pull Requests
* 

#### Changes
*  Wait for DOM to fully load

### DIFF
--- a/ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
+++ b/ehr/test/src/org/labkey/test/tests/ehr/AbstractGenericEHRTest.java
@@ -21,6 +21,7 @@ import org.labkey.remoteapi.CommandResponse;
 import org.labkey.remoteapi.SimplePostCommand;
 import org.labkey.test.Locator;
 import org.labkey.test.Locators;
+import org.labkey.test.WebDriverWrapper;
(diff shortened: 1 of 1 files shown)


labkey-tchad (changes requested):
This is redundant; `beginAt` already does this wait. https://github.com/LabKey/testAutomation/blob/develop/src/org/labkey/test/WebDriverWrapper.java#L2017

This sort of thing is very unlikely to cause a stale element problem. I think the problem here is that some link is navigating instead of opening in a new tab like `AbstractGenericEHRTest.verifyLInk` expects. That causes the elements found at the top of `validatePageLinks` to go stale.

PR #1084, opened 2026-01-05
Drop/Create EHR Index Page

labkey-martyp:
#### Rationale
Initially this page allows dropping and recreating the EHR index on the clinical_observations dataset to resolve an issue introduced in 25.7 (and fixed in 25.11) with recreating indices by name after updating a domain. This creates actions to drop the index manually and recreate it manually after the domain update is complete.

Additionally we'll keep these actions and page as we prepare for migration from SQL Server to Postgres and go through the other EHR indices to see which are still needed. Or if new ones are needed.

#### Changes
* Add new Postgres Migration admin page
* Create actions for dropping and creating the EHR indices.
* Only add clinical_observations index for now.

### DIFF
--- a/ehr/src/org/labkey/ehr/EHRController.java
+++ b/ehr/src/org/labkey/ehr/EHRController.java
@@ -684,6 +684,54 @@ public class EHRController extends SpringActionController
         }
     }
 
+    @RequiresPermission(AdminPermission.class)
+    public static class DropEHRIndicesAction extends MutatingApiAction<Object>
+    {
+        @Override
+        public ApiResponse execute(Object form, BindException errors)
+        {
+            try
+            {
+                List<String> messages = EHRManager.get().dropEHRIndices(getContainer(), getUser());
+                Map<String, Object> response = new HashMap<>();
+                response.put("success", true);
+                response.put("messages", messages);
+                return new ApiSimpleResponse(response);
+            }
+            catch (Exception e)
+            {
+                Map<String, Object> response = new HashMap<>();
+                response.put("success", false);
+                response.put("message", e.getMessage());
+                return new ApiSimpleResponse(response);
+            }
+        }
+    }
+
+    @RequiresPermission(AdminPermission.class)
+    public static class AddEHRIndicesAction extends MutatingApiAction<Object>
+    {
+        @Override
+        public ApiResponse execute(Object form, BindException errors)
+        {
+            try
+            {
+                List<String> messages = EHRManager.get().addEHRIndices(getContainer(), getUser());
+                Map<String, Object> response = new HashMap<>();
+                response.put("success", true);
+                response.put("messages", messages);
+                return new ApiSimpleResponse(response);
+            }
+            catch (Exception e)
+            {
+                Map<String, Object> response = new HashMap<>();
+                response.put("success", false);
+                response.put("message", e.getMessage());
+                return new ApiSimpleResponse(response);
+            }
+        }
+    }
+
     public static class AnimalDetailsForm
     {
         private String[] _animalIds;
--- a/ehr/src/org/labkey/ehr/EHRController.java
+++ b/ehr/src/org/labkey/ehr/EHRController.java
@@ -684,39 +684,55 @@ public class EHRController extends SpringActionController
         }
     }
 
-    @RequiresPermission(AdminPermission.class)
-    public static class DropEHRIndicesAction extends MutatingApiAction<Object>
+    public static class UpdateEHRIndicesForm
     {
-        @Override
-        public ApiResponse execute(Object form, BindException errors)
+        private String _operation;
+
+        public String getOperation()
         {
-            try
-            {
-                List<String> messages = EHRManager.get().dropEHRIndices(getContainer(), getUser());
-                Map<String, Object> response = new HashMap<>();
-                response.put("success", true);
-                response.put("messages", messages);
-                return new ApiSimpleResponse(response);
-            }
-            catch (Exception e)
-            {
-                Map<String, Object> response = new HashMap<>();
-                response.put("success", false);
-                response.put("message", e.getMessage());
-                return new ApiSimpleResponse(response);
-            }
+            return _operation;
+        }
+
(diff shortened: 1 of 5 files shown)


ehr/src/org/labkey/ehr/EHRManager.java
```diff
@@ -1046,12 +1046,151 @@ private boolean doesIndexExist(DbSchema schema, String tableName, String indexNa
...
+
+        TableInfo realTable = StorageProvisioner.createTableInfo(dataset.getDomain());
+        String tableName = dataset.getDomain().getStorageTableName();
+        DbSchema schema = realTable.getSchema();
+
+        String[] indexCols = new String[]{"participantid", "date", "include:taskid,lsid,category,observation,area,remark"};
+        String indexName = getIndexName(schema.getSqlDialect(), tableName, indexCols);
+
+        List<String> cols = Arrays.asList("participantid", "date");
+        String[] includedCols = new String[]{"taskid", "lsid", "category", "observation", "area", "remark"};
```
labkey-jeckels: Why are these column lists hard-coded twice?

ehr/src/org/labkey/ehr/EHRManager.java
```diff
@@ -1046,12 +1046,151 @@ private boolean doesIndexExist(DbSchema schema, String tableName, String indexNa
...
+    }
+
+    /**
+     * Drop indices created for EHR datasets.
+     */
+    public List<String> dropEHRIndices(Container c, User u)
+    {
+        List<String> messages = new ArrayList<>();
+
+        EHRIndexContext ctx = prepareEHRIndexContext(c, messages);
```
labkey-jeckels: This feels over-engineered for a single index and under-engineered for supporting more indices in the future. Consider making `prepareEHRIndexContext()` return a `List` with a single entry for now, and iterate over the list when creating or dropping indices.

ehr/src/org/labkey/ehr/EHRManager.java
```diff
@@ -1046,12 +1046,151 @@ private boolean doesIndexExist(DbSchema schema, String tableName, String indexNa
...
+            return messages;
+        }
+
+        try (DbScope.Transaction transaction = ExperimentService.get().ensureTransaction())
+        {
+            boolean exists = doesIndexExist(ctx.schema, ctx.tableName, ctx.indexName);
+            if (exists)
+            {
+                dropIndex(ctx.schema, ctx.realTable, ctx.indexName, ctx.cols, ctx.dataset.getLabel(), messages);
+                messages.add("Successfully dropped clinical_observations index: " + ctx.indexName);
```
labkey-jeckels: Please make this use the dataset name instead of hard-coding `clinical_observations`

ehr/src/org/labkey/ehr/EHRController.java
```diff
@@ -684,6 +684,54 @@ public ApiResponse execute(RecordDeleteForm form, BindException errors)
...
+                Map<String, Object> response = new HashMap<>();
+                response.put("success", false);
+                response.put("message", e.getMessage());
+                return new ApiSimpleResponse(response);
+            }
+        }
+    }
+
+    @RequiresPermission(AdminPermission.class)
+    public static class AddEHRIndicesAction extends MutatingApiAction<Object>
```
labkey-jeckels: This isn't a huge chunk of code to duplicate across actions, but consider a single action that takes a parameter that distinguishes between `add` and `drop`

ehr/src/org/labkey/ehr/EHRManager.java
```diff
@@ -1046,12 +1046,151 @@ private boolean doesIndexExist(DbSchema schema, String tableName, String indexNa
     private void dropIndex(DbSchema schema, TableInfo realTable, String indexName, List<String> cols, String tableName, List<String> messages)
     {
         messages.add("Dropping index on column(s): " + StringUtils.join(cols, ", ") + " for dataset: " + tableName);
-        String sqlString = "DROP INDEX " + indexName + " ON " + realTable.getSelectName();
+        String sqlString;
+        if (realTable.getSqlDialect().isSqlServer())
```
labkey-jeckels, line 1131: Can we share index dropping code with `ensureDatasetProperyDescriptor()` instead of duplicating?

ehr/src/org/labkey/ehr/EHRManager.java
```diff
@@ -1046,12 +1046,151 @@ private boolean doesIndexExist(DbSchema schema, String tableName, String indexNa
...
+        {
+            boolean exists = doesIndexExist(ctx.schema, ctx.tableName, ctx.indexName);
+            if (exists)
+            {
+                messages.add("Index already exists: " + ctx.indexName);
+            }
+            else
+            {
+                createIndex(ctx.schema, ctx.realTable, ctx.dataset.getLabel(), ctx.indexName, ctx.cols, ctx.includedCols, messages);
+                messages.add("Successfully created clinical_observations index: " + ctx.indexName);
```
labkey-jeckels: Similar note on the hard-coded dataset name

labkey-jeckels (changes requested):

ehr/src/org/labkey/ehr/EHRManager.java
labkey-martyp: I updated the hard coded values in this file and the common index here.

ehr/src/org/labkey/ehr/EHRManager.java
labkey-martyp: Yeah I'm anticipating a lot of refactoring in this file as we look into these indices. I was planning on doing more refactoring at that point but went ahead and set this up more generically for more indices.

ehr/src/org/labkey/ehr/EHRManager.java
labkey-martyp: ensureDatasetProperyDescriptor() was previously using dropIndex in one place and not in another. Fixed.

labkey-jeckels (approved):

PR #1089, opened 2026-01-07
Add supportsNonIdFilters to ehr.reports

labkey-martyp:
#### Rationale
As we convert animal history to react, we'll want to allow filters beyond lists of ids. Not all reports currently support that, so this flag will indicate to animal history how to handle the reports.

#### Changes
* Upgrade script to add supportsNonIdFilters to ehr.reports

### DIFF
--- a/ehr/resources/schemas/dbscripts/postgresql/ehr-26.000-26.001.sql
+++ b/ehr/resources/schemas/dbscripts/postgresql/ehr-26.000-26.001.sql
@@ -1,0 +1,1 @@
+ALTER TABLE ehr.reports ADD supportsNonIdFilters BOOL DEFAULT NULL;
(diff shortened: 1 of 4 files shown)


labkey-jeckels (approved):
How do you plan to handle rows with null values? Are they assumed to be true or false? Would it be more convenient to set them to that value and not allow nulls (maybe not due to existing codepaths that insert rows into the table)?

labkey-martyp:
> How do you plan to handle rows with null values? Are they assumed to be true or false? Would it be more convenient to set them to that value and not allow nulls (maybe not due to existing codepaths that insert rows into the table)?

Yeah I'm just coding it up as null is false. Didn't want to force that update to false everywhere.

PR #1094, opened 2026-01-15
Fix crawl test - postgres migration UI

labkey-martyp:
#### Rationale
Crawl test broken with new postgres migration page

#### Changes
* add to skip list


labkey-bpatel (approved):
Could you please update the PR title to reflect the specific fix? There’s a crawler test that fails intermittently in a different way, so having that distinction would be helpful.

PR #1102, opened 2026-02-14
Allow registering multiple InvoicedItemsProcessingService impls

labkey-adam:
#### Rationale
https://github.com/LabKey/kanban/issues/1577

We need to allow multiple modules to register a `InvoicedItemsProcessingService` impl. Provide a module at registration time and a container at retrieval time. Determine the impl to return based on active modules in the container.

#### Related Pull Requests
* https://github.com/LabKey/wnprc-modules/pull/917
* https://github.com/LabKey/tnprcEHRModules/pull/248

### DIFF
--- a/ehr/resources/schemas/dbscripts/sqlserver/ehr-26.001-26.002.sql
+++ b/ehr/resources/schemas/dbscripts/sqlserver/ehr-26.001-26.002.sql
@@ -1,0 +1,7 @@
+-- ehr.Project.Created and ehr.Project.Modified are NULL on SQL Server but NOT NULL on PostgreSQL. Set the NULL values
+-- and switch the columns to NOT NULL to match PostgreSQL.
+UPDATE ehr.Project SET Created = diCreated WHERE Created IS NULL;
+UPDATE ehr.Project SET Modified = diModified WHERE Modified IS NULL;
+
+ALTER TABLE ehr.Project ALTER COLUMN Created DATETIME NOT NULL;
+ALTER TABLE ehr.Project ALTER COLUMN Modified DATETIME NOT NULL;
--- a/ehr/resources/schemas/dbscripts/sqlserver/ehr-26.001-26.002.sql
+++ b/ehr/resources/schemas/dbscripts/sqlserver/ehr-26.001-26.002.sql
@@ -1,7 +1,7 @@
 -- ehr.Project.Created and ehr.Project.Modified are NULL on SQL Server but NOT NULL on PostgreSQL. Set the NULL values
 -- and switch the columns to NOT NULL to match PostgreSQL.
-UPDATE ehr.Project SET Created = diCreated WHERE Created IS NULL;
-UPDATE ehr.Project SET Modified = diModified WHERE Modified IS NULL;
+UPDATE ehr.Project SET Created = COALESCE(diCreated, GETDATE()) WHERE Created IS NULL;
+UPDATE ehr.Project SET Modified = COALESCE(diModified, GETDATE()) WHERE Modified IS NULL;
 
 ALTER TABLE ehr.Project ALTER COLUMN Created DATETIME NOT NULL;
 ALTER TABLE ehr.Project ALTER COLUMN Modified DATETIME NOT NULL;
--- a/ehr_billing/api-src/org/labkey/api/ehr_billing/pipeline/InvoicedItemsProcessingService.java
+++ b/ehr_billing/api-src/org/labkey/api/ehr_billing/pipeline/InvoicedItemsProcessingService.java
@@ -22,32 +22,45 @@ import org.labkey.api.data.Container;
 import org.labkey.api.data.SimpleFilter;
 import org.labkey.api.data.TableInfo;
 import org.labkey.api.data.TableSelector;
+import org.labkey.api.module.Module;
 import org.labkey.api.pipeline.PipelineJobException;
 import org.labkey.api.query.FieldKey;
 import org.labkey.api.query.QueryService;
 import org.labkey.api.security.User;
-import org.labkey.api.services.ServiceRegistry;
 import org.labkey.api.util.Pair;
 
 import java.util.Collections;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 /**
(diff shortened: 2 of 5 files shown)


ehr_billing/src/org/labkey/ehr_billing/pipeline/BillingTask.java
```diff
@@ -78,11 +78,13 @@
...
 {
     private final static DbSchema EHR_BILLING_SCHEMA = EHR_BillingSchema.getInstance().getSchema();
-    private final static InvoicedItemsProcessingService processingService =  InvoicedItemsProcessingService.get();
+
+    private final InvoicedItemsProcessingService _processingService;
 
     protected BillingTask(Factory factory, PipelineJob job)
     {
         super(factory, job);
+        _processingService = InvoicedItemsProcessingService.get(job.getContainer());
```
labkey-martyp: I think we'll want to use EHR_BillingManager.get().getBillingContainer(getJob().getContainer());

labkey-martyp (commented):
Not sure if you were ready for code review on this. We'll need register calls then in the client repos?

labkey-adam:
> Not sure if you were ready for code review on this. We'll need register calls then in the client repos?

Yes, this PR references PRs in WNPRC and TNPRC that adjust registration. I've assigned you as reviewer to those.

ehr_billing/src/org/labkey/ehr_billing/pipeline/BillingTask.java
labkey-adam: > I think we'll want to use EHR_BillingManager.get().getBillingContainer(getJob().getContainer());

Okay, I'll make that change. Not needed in EHR_BillingController.java?

Does relying on the registration module being active in the billing container make sense to you? IMO, this is more reliable than using folder type (which the old comment had suggested).

labkey-martyp: Yeah probably should use the billing container as described above in the controller too. I don't think it will matter on a production server, but will probably matter on our test servers.

I think using the active modules in that container makes sense for production and test. Production will probably set that module property site wide, but tests will need to set it at the container level.

labkey-adam: Then I should probably just move the EHR_BillingManager.get().getBillingContainer() call into `get()`

labkey-adam: ...except that the interface can't see the manager. So never mind.

ehr/resources/schemas/dbscripts/sqlserver/ehr-26.001-26.002.sql
```diff
@@ -0,0 +1,7 @@
+-- ehr.Project.Created and ehr.Project.Modified are NULL on SQL Server but NOT NULL on PostgreSQL. Set the NULL values
+-- and switch the columns to NOT NULL to match PostgreSQL.
+UPDATE ehr.Project SET Created = diCreated WHERE Created IS NULL;
+UPDATE ehr.Project SET Modified = diModified WHERE Modified IS NULL;
```
labkey-jeckels: We could fall back on `GETDATE()` if the `di` columns are null. Unless we want to hard-fail the upgrade on a server that might not have values as a way of assessing real-world data.

ehr_billing/api-src/org/labkey/api/ehr_billing/pipeline/InvoicedItemsProcessingService.java
```diff
@@ -22,32 +22,45 @@
...
+        REGISTRATION_LIST.add(new Registration(module.getName(), impl));
+    }
+
     @Nullable
-    static InvoicedItemsProcessingService get()
+    static InvoicedItemsProcessingService get(Container billingContainer)
     {
-        return ServiceRegistry.get().getService(InvoicedItemsProcessingService.class);
+        // Return the service implementation based on the registering module being active in the provided container
+        return REGISTRATION_LIST.stream()
```
labkey-jeckels, line 59: In some other EHR contexts, we use a module-dependency ordering to choose between something provided by the core EHR module and a center-specific module.

Here, I believe that the core EHR module would startup first, and therefore have register things first. Then center-specific modules would register. `findFirst()` means we'd resolve the core EHR version, if present. Consider switching to reverse the list or otherwise choose the other version.

labkey-jeckels (approved):

ehr_billing/api-src/org/labkey/api/ehr_billing/pipeline/InvoicedItemsProcessingService.java
labkey-martyp: I assumed we were just registering this in the center specific modules (not in any core EHR modules) so there would be just one service registered in the billing container for each center's project. I believe that would work for the test server with multiple EHR projects.

labkey-martyp (approved):

ehr_billing/api-src/org/labkey/api/ehr_billing/pipeline/InvoicedItemsProcessingService.java
labkey-jeckels: > I assumed we were just registering this in the center specific modules (not in any core EHR modules) so there would be just one service registered in the billing container for each center's project. I believe that would work for the test server with multiple EHR projects.

I believe that's the current state in terms of what needs to be registered, but I'm advocating that we consider being consistent so we don't have to remember which one will win in scenarios like this.

labkey-martyp: Got it. Makes sense.

labkey-adam: Instead of reversing the list or stream every time, registration now uses `addFirst()`

ehr/resources/schemas/dbscripts/sqlserver/ehr-26.001-26.002.sql
labkey-adam: Good idea. I added use of your favorite function, `COALESCE()`.

PR #1112, opened 2026-04-06
Migrate SND to PostgreSQL

labkey-adam:
#### Rationale
We want the SND module to support PostgreSQL

### DIFF
--- a/snd/src/org/labkey/snd/CategoryUserSchema.java
+++ b/snd/src/org/labkey/snd/CategoryUserSchema.java
@@ -38,7 +38,9 @@ public class CategoryUserSchema extends UserSchema
     @Override
     public Set<String> getTableNames()
     {
-        var sql = new SQLFragment("SELECT Description FROM snd.PkgCategories WHERE Active=1 AND Container=").appendValue(getContainer());
+        var sql = new SQLFragment("SELECT Description FROM snd.PkgCategories WHERE Active = ? AND Container = ?")
+            .add(true)
+            .add(getContainer());
         List<String> list = new SqlSelector(getDbSchema(), sql).getArrayList(String.class);
 
         // check for duplicates
@@ -55,8 +57,10 @@ public class CategoryUserSchema extends UserSchema
     @Override
     public @Nullable TableInfo createTable(String name, ContainerFilter cf)
     {
-        var sql = new SQLFragment("SELECT CategoryId, Description FROM snd.PkgCategories WHERE Active=1 AND Container=").appendValue(getContainer())
-                .append(" AND lower(description) = lower(").appendValue(name).append(")");
+        var sql = new SQLFragment("SELECT CategoryId, Description FROM snd.PkgCategories WHERE Active = ? AND Container = ?")
+            .add(true)
+            .add(getContainer())
+            .append(" AND lower(description) = lower(").appendValue(name).append(")");
         Map<String,Object>[] rs = new SqlSelector(getDbSchema(), sql).getMapArray();
         if (rs.length != 1)
             return null;
--- a/snd/module.properties
+++ b/snd/module.properties
@@ -3,4 +3,5 @@ Label: Structured Narrative Datasets
 Description: Dynamically defined hierarchical datasets.
 License: Apache 2.0
 LicenseURL: http://www.apache.org/licenses/LICENSE-2.0
+SupportedDatabases: mssql, pgsql
 ManageVersion: true
--- a/snd/module.properties
+++ b/snd/module.properties
@@ -3,5 +3,4 @@ Label: Structured Narrative Datasets
 Description: Dynamically defined hierarchical datasets.
 License: Apache 2.0
 LicenseURL: http://www.apache.org/licenses/LICENSE-2.0
-SupportedDatabases: mssql
 ManageVersion: true
--- a/snd/resources/schemas/dbscripts/postgresql/snd-0.000-25.000.sql
+++ b/snd/resources/schemas/dbscripts/postgresql/snd-0.000-25.000.sql
@@ -1,0 +1,538 @@
+/*
+ * Copyright (c) 2017 LabKey Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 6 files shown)


labkey-adam:
Here's a summary of every translation decision that Claude made when converting the SQL Server SND script to PostgreSQL:

  ---
  SQL Server → PostgreSQL Translation Notes

  Type mappings

  ┌───────────────────────────────────────────────────┬─────────────────────────────────────────────┐
  │                    SQL Server                     │                 PostgreSQL                  │
  ├───────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ NVARCHAR(n)                                       │ VARCHAR(n)                                  │
  ├───────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ NVARCHAR(MAX)                                     │ TEXT                                        │
  ├───────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ DATETIME                                          │ TIMESTAMP                                   │
  ├───────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ BIT                                               │ BOOLEAN                                     │
  ├───────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ BIT NOT NULL DEFAULT 0                            │ BOOLEAN NOT NULL DEFAULT FALSE              │
  ├───────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ BIT NOT NULL DEFAULT 1                            │ BOOLEAN NOT NULL DEFAULT TRUE               │
  ├───────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ UNIQUEIDENTIFIER NOT NULL DEFAULT newid()/NEWID() │ ENTITYID NOT NULL DEFAULT gen_random_uuid() │
  ├───────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ UNIQUEIDENTIFIER (nullable)                       │ ENTITYID                                    │
  ├───────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ INTEGER IDENTITY / INT IDENTITY(1,1)              │ SERIAL                                      │
  ├───────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ date                                              │ DATE                                        │
  ├───────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ int                                               │ INTEGER                                     │
  ├───────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ LSIDType / LsidType                               │ LSIDType (same custom domain)               │
  ├───────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ ENTITYID, USERID                                  │ unchanged (LabKey custom domains)           │
  └───────────────────────────────────────────────────┴─────────────────────────────────────────────┘

  Syntax changes

  - Removed all GO batch separators (not valid in PostgreSQL).
  - Removed NONCLUSTERED / CLUSTERED index keywords — PostgreSQL has no clustered index concept; all indexes are created
   as plain CREATE INDEX or CREATE UNIQUE INDEX.
  - Removed WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, ...) index storage options.
  - String concatenation + → ||.
  - SPACE(n) → REPEAT(' ', n).
  - CONVERT(VARCHAR(MAX), expr) → expr::VARCHAR.
  - TOP(1) → LIMIT 1.
  - EXEC core.fn_dropifexists → SELECT core.fn_dropifexists (not needed here since this is a fresh install script with
  no pre-existing objects).

  Functions (fGetSuperPkg, fGetProjectItems)

  - SQL Server inline TVF (RETURNS TABLE AS RETURN (...)) → PostgreSQL CREATE OR REPLACE FUNCTION ... RETURNS TABLE
  (...) LANGUAGE sql AS $$ ... $$.
  - WITH ... AS CTE → WITH RECURSIVE ... AS (required in PostgreSQL even when the query is recursive).
  - Return column types explicitly declared.
  - ROW_NUMBER() OVER (...) in the recursive CTE term — PostgreSQL does not permit window functions in the recursive
  member of WITH RECURSIVE. Replaced with COALESCE(sp.SortOrder, 0)::VARCHAR as the path component. This preserves the
  ordering semantics of the original (path segments are zero-padded sort-order values rather than row numbers), and a
  comment in the file explains why.

  Function fGetAllSuperPkgs

  - The SQL Server cursor loop (CURSOR LOCAL FOR ... FETCH ... WHILE ... INSERT INTO) is replaced with a single CROSS
  JOIN LATERAL against snd.fGetSuperPkg, which is idiomatic PostgreSQL and far more efficient than a procedural loop.
  - Corrected the column types in the return table declaration (the original SQL Server script had several wrong types:
  SuperPkgPath INTEGER, Required INTEGER, Active INTEGER, Repeatable INTEGER).

  Trigger (ti_after_events)

  - SQL Server FOR INSERT AS ... UPDATE ... FROM inserted (statement-level, set-based) → PostgreSQL row-level BEFORE
  INSERT trigger.
  - The trigger logic is split into a trigger function (snd.ti_after_events_fn() RETURNS trigger) and a CREATE TRIGGER
  statement (FOR EACH ROW EXECUTE PROCEDURE).
  - Using BEFORE INSERT instead of AFTER INSERT means NEW.QcState can be set directly before the row is written,
  avoiding the extra UPDATE that the SQL Server version required.

snd/src/org/labkey/snd/CategoryUserSchema.java
```diff
@@ -55,8 +57,10 @@ public Set<String> getTableNames()
     @Override
     public @Nullable TableInfo createTable(String name, ContainerFilter cf)
     {
-        var sql = new SQLFragment("SELECT CategoryId, Description FROM snd.PkgCategories WHERE Active=1 AND Container=").appendValue(getContainer())
-                .append(" AND lower(description) = lower(").appendValue(name).append(")");
+        var sql = new SQLFragment("SELECT CategoryId, Description FROM snd.PkgCategories WHERE Active = ? AND Container = ?")
+            .add(true)
```
labkey-martyp, line 61: I'm assuming jdbc will convert this boolean value appropriately to the underlying database

labkey-martyp (approved):
Looks good. As discussed we'll loop back on fGetProjectItems and fGetSuperPkg on PG in future testing/PR.

labkey-martyp (approved):

snd/src/org/labkey/snd/CategoryUserSchema.java
labkey-adam: Yes, that's one (of many) benefit of using parameters

PR #1132, opened 2026-05-12
Avoid checks that will never be satisfied (Strings vs enums)

labkey-jeckels:
#### Rationale
These if() statements will never evaluate true as they're comparing Strings and enums for equality. It's been that way since at least 2013.

#### Changes
* Delete the effectively dead code

### DIFF
--- a/Viral_Load_Assay/src/org/labkey/viral_load_assay/assay/ABI7500ImportMethod.java
+++ b/Viral_Load_Assay/src/org/labkey/viral_load_assay/assay/ABI7500ImportMethod.java
@@ -383,22 +383,6 @@ public class ABI7500ImportMethod extends DefaultVLImportMethod
                         continue;
                     }
 
-                    if (TYPE.Standard.equals(category))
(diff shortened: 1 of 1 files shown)


bbimber:
hi @labkey-jeckels, yes i actually noticed this when i was reviewing your other PR. do you know if this import method is actually in use anymore? To my knowledge WNPRC is the only center that would have used this, and I thought that instrument was no longer in use.

labkey-adam (approved):

labkey-jeckels:
> hi @labkey-jeckels, yes i actually noticed this when i was reviewing your other PR. do you know if this import method is actually in use anymore? To my knowledge WNPRC is the only center that would have used this, and I thought that instrument was no longer in use.

I don't know. @labkey-mohara, if you have a chance to ask WNPRC on the next call, we may be able to complete delete this class. The type used should be captured as a batch property (hidden by default) for Viral Load assay types. The one in question would show up as `ABI 7500`

PR #1133, opened 2026-05-18
Add Delete Empty Tasks admin action to ehr.tasks grid

labkey-martyp:
#### Rationale
Provide an admin-only way to clean up empty `ehr.tasks` rows directly from the tasks grid.

#### Related Pull Requests
* _none_

#### Changes
* Add `DiscardEmptyTasksButton` — a More Actions menu item on the `ehr.tasks` grid, gated on `EHRDataAdminPermission`.
* Add an `EHRController` endpoint that deletes the selected `ehr.tasks` rows after verifying none have referencing rows in `study.studydata`. Tasks with associated study dataset data are rejected with an error; data in other schemas is not checked.


labkey-bpatel (approved):

labkey-martyp (commented):
Admin only UI not needing additional tests.

